Issue #135, opened 2026-04-08
Guide: Coding standards and engineering best practices


blackhathckr:
## Purpose
This project follows production-grade software engineering standards. Every PR will be reviewed against these guidelines. Read this fully, bookmark it, and refer back before submitting any PR.

---

### Consistency
- Use the **existing code style** in the file you are editing. Match indentation, spacing, and patterns.
- TypeScript strict mode is on. No `any` types unless absolutely unavoidable (and commented why).
- Use `const` by default. Use `let` only when reassignment is needed. Never use `var`.
- Use arrow functions for callbacks and inline functions. Use named `function` declarations for top-level/exported functions.

### Naming Conventions
| What | Convention | Example |
|------|-----------|---------|
| Files (components) | kebab-case | `league-settings.tsx` |
| Files (utilities) | kebab-case | `normalize-points.ts` |
| React components | PascalCase | `LeaderboardCard` |
| Functions/variables | camelCase | `getTeamScore`, `isLoading` |
| Constants | UPPER_SNAKE_CASE | `MAX_TEAM_SIZE` |
| Types/Interfaces | PascalCase | `LeagueConfig`, `UserRole` |
| Boolean variables | prefix with is/has/can/should | `isLoading`, `hasAccess`, `canEdit` |
| Event handlers | prefix with handle/on | `handleSubmit`, `onPress` |
| API routes | kebab-case | `/api/leagues/[id]/rest-days` |

### Imports
- Group imports in this order (with a blank line between groups):
  1. External packages (`react`, `next`, `expo`, etc.)
  2. Internal aliases/lib (`@/components`, `@/lib`, `heroui-native`)
  3. Relative imports (`./`, `../`)
- Remove unused imports before committing.

---

### Web App (Next.js)
```
src/
  app/            # Routes (file-based routing)
  components/     # Reusable UI components
  lib/            # Utilities, services, helpers
  config/         # App configuration
```

### Mobile App (Expo)
```
src/
  app/            # Routes (file-based routing via Expo Router)
  components/     # Reusable UI components
  contexts/       # React contexts (auth, theme, drawer)
  lib/            # Utilities, services, helpers
  data/           # Static data (countries, etc.)
  helpers/        # Type definitions, shared hooks
lib/              # HeroUI Native component library (DO NOT EDIT)
```

### Rules
- One component per file. File name matches the default export.
- Co-locate related files. A component's types, styles, and tests go next to it, not in a separate `types/` folder.
[...]

PR #141, opened 2026-04-11
#126 fixed skeleton loading UI for all slow pages

GitGuru-sudo:
## Summary
Fixes an issue where adding an unassigned (unallocated) member to a team would fail with an error.

## Related Issue
Closes #125

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Refactor (no functional change)
- [ ] UI/UX improvement
- [ ] Chore (dependencies, config, CI)
- [ ] Documentation

## Changes Made
- Resolve `NULL` value handling issue in `src/app/api/leagues/[id]/teams/[teamId]/members/route.ts` where the `team_id !== null` check was incorrectly blocking valid assignments.
- Update `POST` handler logic to explicitly permit assignment for members with a `null` `team_id`.
- Fix UI bug in `src/components/teams/add-members-dialog.tsx` where error toasts were displayed for valid selection states.
- Correct error handling logic in `use-league-teams.ts` hook to properly propagate and display backend validation messages to the user.
- Refactor `handleAddSelected` in `AddMembersDialog` to handle sequential API calls more gracefully and provide per-member success/failure feedback.
- Enhance `MutationResult` type in `use-league-teams.ts` to include optional error strings for better UI feedback loops.

## How to Test
1. As a Host or Governor, try adding an unallocated member to a team (should succeed).
2. As a Captain, try adding an unallocated member to your own team (should succeed).
3. Verify that a success toast is shown upon successful addition.
4. Verify that clear error messages are surfaced from the backend if validation fails.

## Screenshots / Recordings
*Not applicable*

## Checklist
- [x] I have tested this locally and it works
- [x] `pnpm build` passes with no errors
- [x] No `console.log` or commented-out code left behind
- [x] My branch is up to date with `develop`
- [x] I have not modified any files outside the scope of this task

Issue #131, opened 2026-04-08
Fix attach workout auto-action in team chat


blackhathckr:
## Bug
"Attach Workout" in team chat auto-attaches a link without letting the user choose what to share. User has zero control over what gets attached.

## Location
- `src/components/messaging/message-input.tsx` — deep link attachment logic (lines 450-483)
- `src/components/messaging/chat-window.tsx`

## Expected Behavior
When user clicks "Attach Workout":
1. Show a selection UI (list of recent workouts or a picker)
2. User selects which workout to share
3. Only the selected workout gets attached to the message

## Current Behavior
Clicking attach auto-sends a link without any user selection step.

## How to Fix
1. Investigate the deep link popover menu items (lines 450-483 in message-input.tsx)
2. Add a selection step between clicking "Attach Workout" and the link being added
3. If this is an edge case or intermittent, document the reproduction steps in the PR

## Acceptance Criteria
- [ ] User can choose which workout to attach
- [ ] No auto-attachment without user confirmation
- [ ] Attached workout shows correctly in the chat message
- [ ] Create PR against `develop` branch

blackhathckr:
Closed — fixed by PR #187 (merged to develop)

PR #142, opened 2026-04-12
Fix workout attach flow to require explicit selection (#135)

ananyactrl:
Fixes the auto-attachment behavior in team chat by requiring explicit workout selection before attaching.

- Deep links remain page-level navigation shortcuts.
- Auto-attachment is fixed: users now explicitly choose a workout before anything attaches.
- Workout picker refreshes on open and shows activity labels when available.

Follow-up (if needed): add record-level deep links with a dedicated effortentry query and link format.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,4 @@
+## [Unreleased] — v0.1.0
+
+### Fixed
+- Fix workout attachments in team chat so users choose a specific recent workout before sharing (#135)
--- a/src/components/messaging/message-bubble.tsx
+++ b/src/components/messaging/message-bubble.tsx
@@ -1,6 +1,6 @@
 'use client';
 
-import { useMemo, useState } from 'react';
+import { useMemo, useState, type ReactNode } from 'react';
 import Link from 'next/link';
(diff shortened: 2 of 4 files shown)


blackhathckr (changes requested):
### Verdict: Request Changes

**Must fix:**
1. **PR template not used** — description is plain text, missing Summary/Related Issue/Type of Change/Checklist sections. Rewrite using the PR template.
2. **Wrong issue reference** — title and branch say #135, but the correct issue is #131. Fix the PR title and description to reference #131; branch name can stay if renaming is inconvenient, but the PR metadata must be corrected.
3. **CHANGELOG.md is being added as a new file in the diff** — this is dangerous. It will overwrite the existing CHANGELOG.md (which has 134 lines of version history) on merge. Do not add a new file — append your entry to the existing CHANGELOG.md on develop.
4. **Removed Challenges, Leaderboard, and Activities attach options** — this is a scope change beyond the workout picker fix. Justify why these were removed or revert them. Removing functionality without discussion is not acceptable.
5. **Breaking change in message-bubble.tsx** — parent_message key changed from `sender_name` to `sender_username`. Verify this matches the actual database column/API response field, otherwise reply bubbles will show undefined.
6. **Vercel check failing** — retrigger after env fix.

**Merge order: 5 of 6 (after #141)**

PR #222, opened 2026-04-26
feature: add 7-phase league lifecycle system

### LINKED ISSUE #170: League lifecycle manager — 7-phase system
## Context
7 phases: Mobilisation, Launch Meet, League Active, Challenges Live, League Closure, Grand Finale, Post-League Archive.

## Scope
Implement the full 7-phase lifecycle system with host dashboard and player views.

## Acceptance Criteria
- [ ] Phase model with transitions in database (league status field extension)
- [ ] Host lifecycle dashboard showing current phase, checklist, days remaining
- [ ] Phase transitions are host-confirmed (never auto-advance, except Trophy to Extended Archive at day 14)
- [ ] Player-facing simplified phase indicator
- [ ] Configurable parameters per league: enrolment period, trial period, challenge frequency

## Dependencies
Depends on League Ended state (#150) and Trophy mode (#161).

dikshaa-012:
## Summary
Implements a DB-backed 7-phase league lifecycle system with host-controlled transitions, a lifecycle dashboard, and a player-facing phase indicator.

## Related Issue
Closes #170 

## Type of Change
- [ ] Bug fix
- [x] New feature
- [x] UI/UX improvement
- [ ] Documentation

## Changes Made
- Add persistent league phase model with 7 lifecycle stages
- Implement host-controlled phase transitions (with Grand Finale → Post-League Archive auto at day 14)
- Build host lifecycle dashboard with phase, checklist, and countdown
- Add player-facing simplified phase indicator
- Introduce configurable league parameters (enrolment period, trial period, challenge frequency)
- Add supporting APIs, hooks, and utilities for phase management

## How to Test
1. Run the latest DB migration for league lifecycle
2. Create or open a league and verify current phase is displayed correctly
3. As host, transition between phases and confirm updates reflect in UI
4. Verify lifecycle dashboard shows checklist and days remaining
5. Check player view to ensure simplified phase indicator is visible
6. Confirm Grand Finale transitions to Post-League Archive after 14 days
7. Validate configurable parameters are reflected in lifecycle behavior

## Checklist


blackhathckr (changes requested):
Good architecture — clean state machine, proper legacy mapping, backward-compatible status field. But several issues need fixing:

### CRITICAL

**1. Missing RLS on `league_phase_history` table**
`supabase/migrations/20260426_phase_lifecycle.sql` — Every table in the project has RLS enabled (see `20241214_complete_rls.sql`). The new table has none. Add:
```sql
ALTER TABLE league_phase_history ENABLE ROW LEVEL SECURITY;
```
Plus appropriate read policies for league members.

**2. GET `/api/leagues/[id]/phase` exposes host-only data to all members**
`src/app/api/leagues/[id]/phase/route.ts:22-37` — Returns full lifecycle info (history, config, checklist, `is_host_only`, `is_archive_due`) to any league member. Regular players should only get `phase`, `phase_label`, `phase_description`, `days_remaining`. Either split the response by role or filter fields.

**3. Empty import — lint/build failure**
`src/app/api/user/leagues/route.ts:4`:
```typescript
import {} from '@/lib/services/leagues';
```
Remove this no-op import.

### HIGH

**4. Dead phase checks in download button**
`src/components/leagues/download-report-button.tsx` — `leaguePhase === 'completed'` and `leaguePhase === 'ended'` will never match any `LeaguePhase` value. Only `'post_league_archive'` is valid. Remove dead checks; add `'grand_finale'` if downloads should work during that phase.

**5. Unused `phase` prop in `ActionCardsProps`**
`src/components/league/dashboard/action-cards.tsx` — `phase?: string` added to interface but never consumed. Either remove it or migrate the existing `league.status` checks to use it.

### MEDIUM

**6. `deriveLeagueStatus` and `persistLeagueStatusIfNeeded` are now dead code**
`src/lib/services/leagues.ts:61-143` — No longer called after this diff. Remove or deprecate.

**7. `LEAGUE_PHASES` and `PHASE_ORDER` are duplicate arrays**
`src/lib/utils/league-phases.ts` — Identical arrays. Remove one, alias the other.

**8. Auto-archive runs on every league read**
`ensureAutomaticLeaguePhaseTransition` is called inside `getLeagueById` — every API call that reads a league triggers an archive check with extra DB reads. Consider moving to a cron job or at least document this intentional side-effect.

blackhathckr (changes requested):
test

blackhathckr (changes requested):
6/8 previous items addressed. 3 issues remain. (1) STILL OPEN: deriveLeagueStatus and persistLeagueStatusIfNeeded function definitions in src/lib/services/leagues.ts:74-143 are dead code -- all call sites removed but definitions remain. Delete them. (2) NEW: Type mismatch in src/components/league/dashboard/header-actions.tsx -- leaguePhase typed as string but DownloadReportButton/DownloadCertificateButton expect LeaguePhase. Will cause TS compile error. Import and use LeaguePhase type. (3) NEW: Dead userId prop in HeaderActionsProps -- no longer passed to any child after download button changes. Remove from interface, destructuring, and call site in src/app/(app)/leagues/[id]/page.tsx. Minor: PHASE_ORDER in league-phases.ts exported but never used. Migration file missing trailing newline.

blackhathckr (changes requested):
Good work fixing all 11 items from the previous reviews. However, there's a runtime-breaking bug:

## BLOCKER

### `params` must be `Promise<>` and awaited (Next.js 16)
**`src/app/api/leagues/[id]/phase/route.ts`** — Both GET and POST handlers declare `{ params }: { params: { id: string } }` and access `params.id` synchronously. This project is on Next.js 16 where route `params` is a `Promise` and must be awaited. As-is, `params.id` will be `undefined` at runtime, causing every call to this endpoint to fail.

Fix both handlers:
```typescript
// GET
export async function GET(
  request: NextRequest,
  { params }: { params: Promise<{ id: string }> },
) {
  const { id: leagueId } = await params;
  ...
}

// POST — same pattern
```

Also fix the catch block in GET which references `params.id` directly — use the already-destructured variable instead.

## WARNINGS (address if possible)

1. **Race condition in phase transition** — `updateLeaguePhase` in `league-lifecycle.ts` does not include `WHERE phase = currentPhase` in the UPDATE. Add `.eq('phase', params.fromPhase)` to prevent concurrent phase skip.

2. **`as any` casts** — `user/leagues/route.ts` uses `(league as any)?.enrolment_period_days` for lifecycle columns. Consider typing the select query result.

3. **`HostLifecycleDashboard` missing null guard** — `phaseInfo.checklist.filter(...)` will throw if a non-host somehow reaches this component. Use optional chaining: `phaseInfo.checklist?.filter(...)`.

## NITs
- `sourceLeagueId` in `getLeagueById` is always equal to `leagueId` — remove the redundant variable.
- Two separate `new Date().toISOString()` calls in `launchLeague` — use a single `const nowIso`.
- `league-info-section.tsx` uses `phase?: string` instead of `LeaguePhase` type.

Fix the BLOCKER (one-line change per handler) and this is ready to merge.

blackhathckr (changes requested):
Good progress — the Next.js 16 params blocker is fixed, race condition has optimistic locking now, and `as any` casts are cleaned up.

## BUG — Incomplete null guard

### `host-lifecycle-dashboard.tsx:80`

```ts
(phaseInfo.checklist?.filter((item) => item.completed).length /
    phaseInfo.checklist.length) * 100 || 0,
```

The numerator has `?.` but the **denominator** (`phaseInfo.checklist.length`) does **not**. If `checklist` is `null`/`undefined`, the denominator throws `TypeError: Cannot read properties of undefined (reading 'length')`.

**Fix:**
```ts
((phaseInfo.checklist?.filter((item) => item.completed).length ?? 0) /
    (phaseInfo.checklist?.length || 1)) * 100,
```

One-line fix, then this PR is good to go.

blackhathckr (changes requested):
Null guard fix is correct ✓. Three remaining issues:

## 1. Race condition in `updateLeaguePhase` (src/lib/services/league-lifecycle.ts)

The optimistic lock (`eq('phase', fromPhase)`) can fail silently — Supabase `.update()` returns `{ error: null }` even when 0 rows match. But `insertPhaseHistory` still runs, creating an orphaned history entry for a transition that never happened.

**Fix:** Add `.select()` to the update and check `data.length === 0`:
```ts
const { data, error } = await supabase
  .from('leagues')
  .update({ phase: params.toPhase, ... })
  .eq('league_id', params.leagueId)
  .eq('phase', params.fromPhase)
  .select();

if (!data || data.length === 0) {
  return { error: 'Phase was already changed by another request' };
}
```

## 2. Missing null guard for `league.phase` in updateLeague/deleteLeague (src/lib/services/leagues.ts)

`league.phase` is used directly in `.includes(league.phase)` and `league.phase !== 'mobilisation'`. Pre-migration rows with `phase = null` would:
- **updateLeague**: `.includes(undefined)` → returns false → `allowedUpdates` stays empty → silent no-op for settings updates
- **deleteLeague**: `undefined !== 'mobilisation'` → true → blocks deletion of draft leagues that haven't been migrated

**Fix:** Use `normalizeLeaguePhase(league.phase, league.status)` or add `|| ''` guard.

## 3. Inconsistent phase normalization in leaderboard/page.tsx

`league?.phase` is used raw without `normalizeLeaguePhase()`, unlike other call sites. Should normalize for consistency with pre-migration data.

---

Issues 1 and 2 are real bugs. Fix those and this is good to merge.

blackhathckr (approved):
All three issues fixed correctly. Race condition guarded with .select() + length check, null phase handled via normalizeLeaguePhase in both leagues.ts and leaderboard. LGTM — merge when ready.

PR #233, opened 2026-04-28
#149 - PWA login issue

### LINKED ISSUE #149: Fix PWA auto-login redirect for logged-in users
## Context
PWA installed users are not auto-logged-in on revisit. They see the login page despite having an active session.

## Scope
Fix session persistence and redirect logic for PWA users.

## Acceptance Criteria
- [ ] Users with valid session tokens are redirected to dashboard on PWA launch
- [ ] Session persistence works across PWA close/reopen
- [ ] No redirect loop on expired sessions (graceful logout)

## Dependencies
None.

dhjr:
## Summary

Fixed PWA auto-login and session persistence issues by updating the manifest start URL, implementing root-level redirection for authenticated users, and optimizing session refetching behavior.

## Related Issue

Closes #149 

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] UI/UX improvement
- [ ] Documentation

## Changes Made

- Manifest Update: Changed start_url from / to /dashboard in manifest.webmanifest to ensure the PWA launches directly into the application.
- Root Redirection: Added server-side redirection in src/app/page.tsx using getServerSession to automatically send logged-in users to the dashboard.
- Session Optimization: Set refetchOnWindowFocus={true} in AuthProvider to prevent unnecessary session checks and potential race conditions when resuming the PWA from the background.
- Redirection Reliability: Verified and refined redirection logic in AppLayout and LoginPage to ensure a smooth, loop-free experience for both authenticated and unauthenticated users.

## How to Test

1. PWA Launch: Install the PWA and ensure that opening it takes you directly to the dashboard if logged in, or the login page if not.
2. Root Access: Visit the root URL (/) while logged in and verify that you are immediately redirected to /dashboard.
3. Session Persistence: Close the PWA and reopen it multiple times to ensure the session remains active and you land on the dashboard without being forced to the login page.
4. Logout Flow: Log out and ensure you are correctly redirected to the landing page or login screen without any redirect loops.

## Checklist


blackhathckr (changes requested):
## Review: Requesting Changes

The overall approach is sound (manifest start_url + root redirect), but two issues need to be addressed:

### 1. Performance regression on the landing page (Critical)

**File:** `src/app/page.tsx`

Making `RootPage` async with `getServerSession()` forces the entire root `/` page into dynamic rendering. This is the public marketing/corporate landing page — it should remain statically generated for performance (fast TTFB, CDN caching). Every unauthenticated visitor will now hit the server on every page load.

**Suggested fix:** Use Next.js middleware to check the session and redirect only for `/`:

```ts
// src/middleware.ts
import { getToken } from 'next-auth/jwt';
import { NextResponse } from 'next/server';
import type { NextRequest } from 'next/server';

export async function middleware(req: NextRequest) {
  if (req.nextUrl.pathname === '/') {
    const token = await getToken({ req, secret: process.env.NEXTAUTH_SECRET });
    if (token) {
      return NextResponse.redirect(new URL('/dashboard', req.url));
    }
  }
  return NextResponse.next();
}

export const config = { matcher: ['/'] };
```

This keeps the landing page statically renderable for unauthenticated users while still redirecting logged-in users.

### 2. `refetchOnWindowFocus={false}` weakens session expiry detection

**File:** `src/components/auth/auth-provider.tsx`, line 45

Setting `refetchOnWindowFocus={false}` means if a user's session expires while the PWA is backgrounded, the app won't detect it when resumed. The user will see stale UI and get 401 errors on API calls until a full navigation. The original `true` value ensures session validity is checked when the app regains focus.

If there's a specific race condition being fixed, please document it. Otherwise revert this change.

### Minor

- `src/app/page.tsx` — `authOptions as any` is redundant since `authOptions` is already typed as `any`. Remove the cast.

---

The manifest `start_url` change is good. Please address the above and re-request review.

blackhathckr (changes requested):
Re-review: Requesting Changes. See details below.

blackhathckr:
Previous feedback addressed: root redirect moved to middleware, refetchOnWindowFocus reverted to true. But new bug introduced. When unauthenticated user visits /, token is null. Code enters validation block because isRoot=true. First two checks skip. Falls through to (token as any).needsProfileCompletion which throws TypeError on null. Catch block redirects to /login, breaking landing page. Fix: add if (!token and isRoot) return res after second check.

blackhathckr:
Minor issues: PR description still says refetchOnWindowFocus=false but code keeps it true - update description. CHANGELOG says optimizing session refetching behavior referring to reverted change - reword. Missing space before (#149) in CHANGELOG entry.

machinist-oss (approved):
Null token bug fixed correctly. Root path logic is sound: authenticated root redirects to dashboard, unauthenticated root passes through to landing page, unauthenticated protected redirects to login. Duplicate /dashboard cleaned up. CHANGELOG updated. Minor: PR description still mentions optimizing session refetching but that was reverted - update the description. Not a blocker. LGTM.

PR #244, opened 2026-05-03
#148 - points normalization

dhjr:
## Summary

Fixes the points normalisation calculation to only apply the team-size multiplier to individual points and individual-level challenges. This prevents smaller teams from receiving massive, unfair multiplier advantages for fixed team challenges.

## Related Issue

Closes #148 

## Type of Change

- [x] Bug fix
- [ ] New feature
- [x] Refactor (no functional change)
- [ ] UI/UX improvement
- [ ] Chore (dependencies, config, CI)
- [ ] Documentation

## Changes Made

- Modified `src/lib/services/leaderboard-logic.ts` to track `individual_challenge_points` and `fixed_team_points` separately instead of bundling them together.
- Updated `src/hooks/use-league-leaderboard.ts` to only apply the `(max_team_size / team_member_count)` normalisation multiplier to base `points` and `individual_challenge_points`, leaving `fixed_team_points` (like team challenges and special challenges) untouched.
- Created a new Jest test suite at `src/lib/utils/normalization.test.ts` to mathematically verify the fairness of the fix across edge cases involving uneven team sizes (e.g. 5, 7, and 8 members).
- Added a `CHANGELOG.md` entry under the `v2.6.0` fixed section.

## How to Test

1. Create or view a league that has the "Normalise Points by Team Size" setting enabled.
2. Observe teams with uneven sizes (e.g., a team of 5 members and a team of 8 members).
3. Have both teams complete a Team-Level Challenge that awards a fixed amount of points (e.g., 50 points).
4. Verify on the leaderboard that both teams receive exactly 50 points for the challenge. (Before this fix, the smaller team would receive an inflated 80 points due to the 1.6x multiplier).
5. Run the new unit test locally using `npm run test -- src/lib/utils/normalization.test.ts` to ensure it passes.

## Checklist
- [x] I have tested this locally and it works
- [x] `pnpm build` passes with no errors
- [x] No `console.log` or commented-out code left behind
- [x] My branch is up to date with `develop`
- [x] I have not modified any files outside the scope of this task

PR #245, opened 2026-05-03
feat: implement tiered leaderboard visibility with configurable rank …

### LINKED ISSUE #160: Low-performer motivation logic — tiered rank visibility
## Context
Bottom 30% seeing their low rank is demotivating. Tiered approach from 8 April dev meeting.

## Scope
Implement tiered rank visibility so low performers see encouragement instead of discouraging rank numbers.

## Acceptance Criteria
- [ ] Top 20%: see exact rank and position
- [ ] Middle 40%: see rank band/range + motivational nudge
- [ ] Bottom 30%: do NOT see rank, show encouragement message instead
- [ ] Thresholds configurable per league by host
- [ ] AI nudges reference appropriate tier messaging

## Dependencies
None.

dhjr:
## Summary

This PR implements a tiered rank visibility system for the leaderboard to keep low performers motivated. Instead of showing discouraging exact ranks to everyone, hosts can configure the leaderboard to only show exact ranks to top performers, while middle performers see a rank band (e.g., "Top 60%") and bottom performers see encouraging messages.

## Related Issue

Closes #160 

## Type of Change

- [ ] Bug fix
- [x] New feature
- [x] UI/UX improvement
- [ ] Documentation

## Changes Made
- **Database & API:** Added `tiered_rank_enabled` and `tiered_rank_config` to the `leagues` table (via new SQL migration) and updated the `PATCH /api/leagues/[id]` endpoint to support them.
- **Settings UI:** Added a new "Tiered Leaderboard Visibility" configuration block in the League Settings page under the "Point Normalization" section, allowing hosts to define custom percentage splits (default: 20% Top, 50% Middle, 30% Bottom).
- **Leaderboard Hooks:** Updated `useLeagueLeaderboard` to compute `rank_tier`, `rank_band`, and a personalized `motivational_nudge` for both teams and individuals during normalization.
- **Leaderboard Tables:** Modified `RankBadge` in `LeagueIndividualsTable` and `LeagueTeamsTable` to conditionally hide exact ranks for middle and bottom tiers, replacing them with formatted text/icons and motivational tooltips.
- **AI Coach Integration:** Updated `ai-league-manager` and `ai-nudges` services to inject a team's `rank_tier` into the Mistral prompt context, ensuring AI-generated messages adapt their tone appropriately for struggling teams.

## How to Test

1. Run the new database migration (`supabase/migrations/20260504_tiered_rank_visibility.sql`) in your Supabase SQL editor.
2. Navigate to **League Settings** as a Host and toggle **Tiered Leaderboard Visibility** on.
3. Keep the default settings (Top 20%, Middle 50%, Bottom 30%) or adjust them, and hit **Save Changes**.
4. Go to the **Leaderboard** and observe the `#` column for both Individuals and Teams. Top performers should show an exact rank. Middle performers should show a band (e.g., "Top 50%"). Bottom performers should show an Activity icon.
5. Hover over the Middle and Bottom tier badges to ensure the motivational tooltip appears.

## Checklist


blackhathckr (changes requested):
## PR #245 Review: Tiered Leaderboard Visibility

I have reviewed all 11 changed files in detail. The feature concept is solid and well-structured, but there are several issues that need to be fixed before merge — two of them are blockers.

---

### MUST-FIX (Blockers)

**1. `tiered_rank_enabled` / `tiered_rank_config` are NOT in `LeagueInput` — updates silently dropped for active leagues**

The `updateLeague()` service in `src/lib/services/leagues.ts` accepts `Partial<LeagueInput>`, but `LeagueInput` (line 20-42) does not include `tiered_rank_enabled` or `tiered_rank_config`. This means:

- For **draft** leagues, the fields pass through because `Object.assign(allowedUpdates, data)` copies everything. However, you lose TypeScript safety — the fields are passed as `any`.
- For **active/launched** leagues, the fields are **silently dropped** because the allowlist at lines 449-491 does not include `tiered_rank_enabled` or `tiered_rank_config`. A host toggling tiered visibility on an active league will see no error, but the setting will never be saved.

**Fix:** Add both fields to `LeagueInput` in `src/lib/services/leagues.ts`, AND add them to the non-draft allowlist (alongside the other "always editable" settings like `cross_team_visibility`).

**2. No validation that top + middle + bottom = 100**

The Zod schema in `src/app/api/leagues/[id]/route.ts` validates each field individually (`z.number().int().min(0).max(100)`) but never validates that the three percentages sum to 100. A host could submit `top: 80, middle: 80, bottom: 80` (sum = 240) or `top: 10, middle: 10, bottom: 10` (sum = 30) and it would pass validation, producing broken tier cutoffs.

**Fix:** Add a `.refine()` on the `tiered_rank_config` object:
```ts
.refine(
  (cfg) => cfg.top_percent + cfg.middle_percent + cfg.bottom_percent === 100,
  { message: 'Tier percentages must sum to 100' }
)
```
Also add matching client-side validation in the settings page before save.

---

### SHOULD-FIX (Important but not blocking)

**3. Settings UI: no client-side sum validation or feedback**

The three percentage inputs allow any value 0-100 each. There is no UI indicator (e.g., red text, disabled Save) when they do not sum to 100. Users will get confused.

**Fix:** Add a validation message like `"Must sum to 100 (currently: ${sum})"` below the three inputs, and disable Save when invalid.

**4. Edge case: all tiers at 0% or a single tier at 100%**

If a host sets `top: 100, middle: 0, bottom: 0`, every team gets `rank_tier: 'top'` (which is fine). But `top: 0, middle: 100, bottom: 0` means `topCutoff = Math.max(1, 0) = 1`, so the first team still shows as "top" due to the `Math.max(1, ...)` guard, which is inconsistent with the intended config.

Consider documenting or enforcing a minimum `top_percent >= 1` or adjusting the logic so that `top_percent: 0` actually means zero teams in top tier.

**5. `assignRankTiers` uses array index, not `.rank` — fragile if items are not perfectly sorted by rank**

In `use-league-leaderboard.ts`, `assignRankTiers` uses `index` (the array position) to determine which tier an item falls into:
```ts
if (index < topCutoff) { rank_tier = 'top'; }
```
But the items may already have a `.rank` field that could differ from array index (e.g., after normalization, ties, etc.). Using `item.rank - 1` instead of `index` would be more robust. Or at least add a comment that this function assumes the array is pre-sorted by rank.

**6. `rank_band` calculation can produce "Top 100%" for middle-tier items**

```ts
const bandPct = Math.ceil(((index + 1) / total) * 10) * 10;
rank_band = `Top ${bandPct}%`;
```
For the last item in the middle tier, `bandPct` could be 80 or 90, which is fine. But for an item near the middle/bottom boundary, `bandPct` could reach 100 — showing "Top 100%" as a rank band is misleading.

**Fix:** Cap it: `const bandPct = Math.min(90, Math.ceil(...))`.

**7. Duplicated `RankBadge` logic across individuals and teams tables**

The `RankBadge` function in `league-individuals-table.tsx` and `league-teams-table.tsx` is now nearly identical (same tooltip logic, same tier handling). Consider extracting a shared `<TieredRankBadge>` component.

**8. AI league manager tiering uses rank, leaderboard hook uses index — inconsistency**

In `ai-league-manager.ts` (line ~420), teams are tiered using `t.rank <= topCutoff` (1-indexed rank), while in `use-league-leaderboard.ts`, tiering uses `index < topCutoff` (0-indexed). Both should be consistent. Currently they produce the same result only if `rank === index + 1` and the arrays are perfectly sorted, which is likely but not guaranteed.

**9. `ai-nudges.ts` — `CategorizedMember.category` type widened from `NudgeCategory` to `string`**

The `category` field in `CategorizedMember` was changed from `NudgeCategory` to `string` to accommodate the `_tier_bottom`/`_tier_middle` suffix. This loses type safety. Consider using a discriminated union or keeping the base category as a separate field alongside `rank_tier`:
```ts
interface CategorizedMember {
  member: MemberSnapshot;
  category: NudgeCategory;
  rank_tier?: 'top' | 'middle' | 'bottom';
}
```
Then compose the prompt using `rank_tier` rather than string-concatenating it into the category.

**10. `ai-league-manager.ts` — `sortedMembers` mutation does not propagate back to `allMembers`**

The member tiering code in `ai-league-manager.ts` does:
```ts
const sortedMembers = [...allMembers].sort(...)
for (let i = 0; i < sortedMembers.length; i++) {
  sortedMembers[i].rank_tier = 'top' | ...
}
```
Since `[...allMembers]` creates a shallow copy, the objects inside are still references — so `rank_tier` mutations on `sortedMembers[i]` DO propagate back to `allMembers`. This works, but only by accident and is confusing. Add a comment clarifying this, or explicitly set `rank_tier` on the original `allMembers` array.

---

### MERGE CONFLICT RISK with PR #244

PR #244 (points normalization) also modifies `src/hooks/use-league-leaderboard.ts`. Specifically:
- #244 adds `individual_challenge_points` and `fixed_team_points` to `TeamRanking` (same hunk where #245 adds `rank_tier`/`rank_band`/`motivational_nudge`)
- #244 changes normalization logic in the hook body

The changes are in adjacent but different locations:
- #244 modifies the `TeamRanking` interface at lines 29-31 (adding two fields after `normalized_points`)
- #245 adds three fields to `TeamRanking` at the same location

**This will produce a merge conflict.** Whichever PR merges first, the second will need a rebase. Since #244 is smaller, I'd suggest merging it first and rebasing this PR.

---

### LOOKS GOOD

- **Migration** (`20260504_tiered_rank_visibility.sql`): Clean `ALTER TABLE ... ADD COLUMN IF NOT EXISTS` with sensible defaults. No data loss risk. JSONB column with default config is appropriate.
- **CHANGELOG.md**: Entry is correctly placed and descriptive.
- **`src/types/admin.ts`**: Types added to all three interfaces (`AdminLeague`, `AdminLeagueCreateInput`, `AdminLeagueUpdateInput`). Clean.
- **`src/lib/ai/types.ts`**: `rank_tier` added to `MemberSnapshot` and `TeamHealthSummary` as optional — correct.
- **`< 3 items` guard**: Good edge-case handling — if there are fewer than 3 teams/individuals, everyone gets `top` tier.
- **Motivational nudges**: Nice touch with the rotating messages. Good that they are deterministic (index % array.length).

---

### VERDICT: CHANGES REQUESTED

Fix blockers #1 and #2 (service layer allowlist + sum validation). Ideally also fix #3 (UI feedback) and #6 (band capping) before merge. The rest can be addressed as follow-ups.

blackhathckr (changes requested):
## Review: Changes Requested (2 must-fix, 5 should-fix)

Good feature design — tiered visibility with configurable percentages, AI tone adaptation, and clean settings UI. Two blockers before merge.

### Must Fix

**1. Settings never persist on active leagues**
`tiered_rank_enabled` and `tiered_rank_config` are missing from the `LeagueInput` type in `src/lib/services/leagues.ts` and from the `updateLeague()` allowlist for non-draft leagues (lines ~449-491). Toggling tiered visibility on an active league silently does nothing — the API returns success but the data is never written.

**2. No validation that tier percentages sum to 100**
The Zod schema in `/src/app/api/leagues/[id]/route.ts` validates each percentage individually (0-100) but never checks `top + middle + bottom === 100`. A host could submit 80/80/80 and get a broken config.

Add: `.refine(data => data.top + data.middle + data.bottom === 100, { message: "Percentages must sum to 100" })`

### Should Fix

**3. No client-side sum validation** — The settings UI lets you save without checking the sum. Add visual feedback or auto-adjust.

**4. `rank_band` edge case** — Can produce "Top 100%" for last-place items. Cap the display.

**5. Duplicated `RankBadge` logic** — Nearly identical in both `league-individuals-table.tsx` and `league-teams-table.tsx`. Extract to a shared component.

**6. Type safety in `ai-nudges.ts`** — `NudgeCategory` was widened from union type to `string`, losing type safety.

**7. Merge conflict with PR #244** — Both modify `TeamRanking` interface in `use-league-leaderboard.ts` at the same location. Merge #244 first, then rebase this PR.

blackhathckr (changes requested):
### Changes Requested

1. **Settings page conflict** — This PR modifies `settings/page.tsx` which was just changed on develop (route-bound host role check). Rebase onto latest develop and resolve conflicts.

2. **No validation that percentages sum to 100** — User can enter top=20, middle=50, bottom=50 = 120%. Add validation in the settings UI that enforces `top + middle + bottom === 100` before saving.

3. **Fragile icon imports** — `league-teams-table.tsx` imports from deep paths like `lucide-react/dist/esm/icons/activity`. This will break on lucide version updates. Use `import { Activity } from 'lucide-react'` instead.

blackhathckr (approved):
All requested changes addressed:
- Percentage validation enforces sum = 100%
- Lucide imports fixed to standard pattern
- No merge conflicts

Approved.

Issue #159, opened 2026-04-13
Engagement metrics tracking engine


blackhathckr:
## Context
Track 10-20 engagement metrics per player for future sponsor/corporate dashboard. From 8 April meeting.

## Scope
Backend instrumentation to capture player engagement events. No user-facing UI yet.

## Acceptance Criteria
- [ ] Track: logins, activity logs, leaderboard views, chat messages sent, challenge participations, AI coach interactions, profile views, streak days, rest days used, time-in-app
- [ ] Store in analytics table (new or existing)
- [ ] Data queryable for dashboard consumption in later milestones
- [ ] No user-facing UI yet — backend instrumentation only

## Dependencies
None.

blackhathckr:
Implemented in PR #241 (merged to develop). Backend engagement tracking engine.

PR #267, opened 2026-05-09
Feat/173 engagement dashboard

### LINKED ISSUE #173: Engagement dashboard beta
## Context
Consumes data from engagement metrics tracking (#159). Host-facing dashboard for engagement visibility.

## Scope
Build a basic engagement dashboard for hosts with key metrics and visualizations.

## Acceptance Criteria
- [ ] Host-facing dashboard showing key engagement metrics
- [ ] Participation rate, daily active users, streak distribution
- [ ] At-risk player list (gap > 2 days)
- [ ] Basic visualizations (charts/graphs)
- [ ] Exportable summary for corporate/HR reporting

## Dependencies
Depends on #159 (engagement metrics tracking engine).

ananyactrl:
## Summary

Engagement dashboard for hosts and governors — key metrics, visualizations, at-risk player detection, and CSV export. Consumes data from the engagement tracking engine (#159).

## Related Issue

Closes #173

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [x] UI/UX improvement
- [ ] Chore
- [ ] Documentation

## Changes Made

**New files:**
- `page.tsx` — server component, role-gates to host/governor only
- `engagement-client.tsx` — full dashboard UI with 6 zones
- `route.ts` — analytics API, parallel queries, no caching

**Modified files:**
- `sidebar-config.ts` — Engagement link added to host/governor nav
- `league-context.tsx` — fixed role downgrade bug for hosts/governors
- `admin/leagues/[leagueId]/memberships/route.ts` — governor access, batch queries
- `admin/leagues/[leagueId]/memberships/[membershipId]/route.ts` — governor access, body-spoofing fix

**6 Dashboard Zones:**
1. Header with league name, Host/Governor badge, Export CSV, Refresh
2. 4 metric cards — Participation Rate, Active (7d), At-Risk, Total Events
3. Daily Active Users bar chart (14 days)
4. Event Breakdown + Streak Distribution
5. At-Risk Players table — inactive >2 days, Nudge tooltip
6. Top Engagers (30d)

## How to Test

1. Host login — Engagement link in sidebar, dashboard loads
2. Governor login — Engagement link visible, no 403
3. Player/captain — redirect to league page, not 404
4. All 6 zones render with real data
5. CSV export — correct column alignment in spreadsheet
6. At-risk table — inactive >2 days, "30d+" for null dates
7. Nudge tooltip on hover (desktop + mobile)
8. Refresh button — loading skeleton, data re-fetches
9. `pnpm build` — zero errors

## Checklist

- [x] I have tested this locally
- [x] `pnpm build` passes
- [x] No `console.log` left behind
- [x] Branch up to date with `develop`


machinist-oss (changes requested):
## Review: Feat/173 Engagement Dashboard

Good work on the dashboard UI and the parallel query design in the engagement API. That said, there are several issues that need to be addressed before merge.

---

### BLOCKER

**B1. Unbounded queries in the engagement API route will hit Supabase default 1000-row limit and silently return truncated data.**

`src/app/api/leagues/[id]/engagement/route.ts`

Several queries fetch all rows without pagination or a `.limit()`:
- Query 2 (active users last 7 days) — fine for small leagues, but a league with 500 members logging daily creates thousands of rows.
- Query 3 (daily events last 14 days) — fetches every event row for 14 days with `user_id, event_timestamp`. For a 200-member league logging 5 events/day, that is 14,000 rows — well over the Supabase default 1000-row cap.
- Query 4 (event breakdown last 30 days) — same issue, even worse window.
- Query 5b (at-risk activity, ALL TIME, no date filter) — unbounded. This will grow forever.
- Query 7 (top engagers last 30 days) — same as query 4.

Supabase PostgREST silently truncates at the server's `max-rows` setting (default 1000). This means the dashboard will show wrong data for any league with moderate activity and the host will never know.

**Fix:** Either (a) move these aggregations into Postgres views / RPC functions so counting happens server-side, or (b) at minimum add explicit `.limit()` calls with pagination or use `{ count: 'exact', head: true }` patterns where you only need counts, not rows. Query 5b must at least be scoped with a date filter or use an RPC that returns `MAX(event_timestamp) GROUP BY user_id`.

**B2. Governor can now delete any league member including hosts — missing privilege escalation guard.**

`src/app/api/admin/leagues/[leagueId]/memberships/[membershipId]/route.ts`

Both `PATCH` and `DELETE` now grant governors the same power as hosts (`callerRole === 'governor'`). There is no check preventing a governor from:
- Promoting themselves to host (PATCH with `role: 'host'`).
- Removing the host from the league (DELETE on the host's membership).

A governor should not be able to modify the host's role or remove the host. Add a guard: if the target `userId === league.created_by`, reject with 403 ("Cannot modify the league creator").

---

### Major

**M1. `as any` casts throughout — type safety is lost.**

Multiple files (`memberships/route.ts`, `memberships/[membershipId]/route.ts`, `engagement/route.ts`) use `(x as any)` to access nested Supabase join fields like `.roles?.role_name` and `.users?.username`. Define local types for the query return shapes so the casts are contained.

**M2. Engagement API fetches raw rows to count in JS when Supabase supports server-side aggregation.**

Even if the 1000-row cap is addressed, pulling thousands of raw event rows to the Node.js process just to `Map.set()` and count is wasteful. The event breakdown (query 4), top engagers (query 7), and daily active users (query 3) should be SQL-level aggregations (e.g., Supabase RPC with `GROUP BY`).

**M3. The `atRiskActivityResult` query (5b) has no date filter and fetches ALL activity_log events ever, ordered descending.**

This is an unbounded full-table scan per league that will get slower over time. Since you only need the most recent event per user, this should be an RPC: `SELECT user_id, MAX(event_timestamp) FROM player_engagement_events WHERE league_id = $1 AND event_type = 'activity_log' GROUP BY user_id`.

---

### Minor

- m1: `viewerRole` prop passed to `EngagementClient` but only used for badge display — document future plans or remove
- m2: Hardcoded hex fallback colors in `chartColors` state won't adapt to theme changes
- m3: CSV export `csvCell()` handles double-quote escaping but not newlines
- m4: `let streakCounts` should be `const` — the object reference is never reassigned

### Nits

- `text-[13px]` — use Tailwind's `text-xs` or `text-sm` instead of arbitrary values

---

### Verdict: REQUEST CHANGES

B1 (unbounded queries / silent truncation) and B2 (governor privilege escalation) must be fixed before merge.

blackhathckr (changes requested):
## R2 Re-review — Engagement Dashboard (#267)

### Previous issues — status table

| ID | Severity | Summary | Status |
|----|----------|---------|--------|
| B1 | BLOCKER | Unbounded queries in engagement API (5 queries fetch all rows, hit 1000-row cap) | **FIXED** — Queries 2/3/4/5b/7 now use Postgres RPCs with server-side aggregation |
| B2 | BLOCKER | Governor privilege escalation — governor can promote self to host or delete host | **FIXED** — Both PATCH and DELETE now check `userId === league?.created_by` and reject with 403 |
| M1 | Major | `as any` casts for Supabase join fields | **FIXED** — Replaced with typed local interfaces (`LeagueRoleRow`, `MemberRow`, `TeamMemberRow`, `MemberUsernameRow`) |
| M2 | Major | Engagement API fetches raw rows to count in JS | **FIXED** — RPCs handle aggregation server-side |
| M3 | Major | `atRiskActivityResult` (query 5b) has no date filter — full table scan | **FIXED** — Moved to `get_last_activity_per_user` RPC (server-side `MAX(event_timestamp) GROUP BY user_id`) |
| M4 | Major | `let streakCounts` should be `const` | **FIXED** — Now `const streakCounts = { ... }` |
| m1 | Minor | `viewerRole` prop passed but only used for badge display | **FIXED** — JSDoc comment added documenting intent for future role-scoped feature gating |
| m2 | Minor | Hardcoded hex fallback colors won't adapt to theme | **FIXED** — CSS custom properties resolved via `getComputedStyle` with `colorsReady` gate; Tailwind class maps for progress bars/streaks |
| m3 | Minor | CSV `csvCell()` doesn't handle newlines | **FIXED** — Now strips `\r?\n` before quoting |
| m4 | Minor | `_request` unused parameter prefix inconsistency | **FIXED** — Consistently `_request` when unused, `request` when used |
| n1 | Nit | `text-[13px]` arbitrary value | **FIXED** — Removed; though many other arbitrary values remain (`text-[10px]`, `text-[36px]`, etc.) these are design-spec values without standard Tailwind equivalents — acceptable |

All R1 issues are resolved. Good work on the refactors — the batch query approach in memberships/route.ts is a significant improvement over the per-member loop.

---

### NEW issues found in R2

#### NEW-B1 (BLOCKER): RPC functions not defined — migration missing

The engagement route calls 4 Supabase RPCs:
- `get_active_user_count`
- `get_daily_active_users`
- `get_event_breakdown`
- `get_last_activity_per_user`
- `get_top_engagers`

**None of these are defined in any SQL migration file.** The existing migration (`20260430_player_engagement_analytics.sql`) only creates the `player_engagement_events` table and indexes. Without these RPCs deployed to Supabase, every `supabase.rpc(...)` call will return an error and the entire engagement page will fail with a 500.

**Fix:** Add a new migration file (e.g., `20260510_engagement_rpcs.sql`) that creates all 5 functions with `CREATE OR REPLACE FUNCTION ... LANGUAGE sql/plpgsql`. Include this migration in the PR.

#### NEW-M1 (Major): effortentry streak query has no `.limit()` — silent 1000-row truncation

```ts
const { data: effortEntries } = await supabase
  .from('effortentry')
  .select('league_member_id, date')
  .in('league_member_id', memberIds)
  .eq('status', 'approved')
  .gte('date', sixtyDaysAgo.toISOString().slice(0, 10))
  .order('date', { ascending: false });
```

This query fetches raw rows for all league members over 60 days. For a league with 50+ members averaging 1 entry/day, that's 3000+ rows — Supabase will silently truncate to 1000, producing **incorrect streak calculations** for some members.

**Fix options (pick one):**
1. Move streak computation to an RPC (consistent with the other fixes).
2. Add `.limit(10000)` to override the default 1000-row cap (quick fix, but fragile).
3. Paginate with range queries.

Option 1 is strongly preferred for consistency.

---

### Verdict: **REQUEST_CHANGES**

NEW-B1 is a deploy-breaker — the RPCs don't exist. NEW-M1 will silently corrupt streak data for active leagues. Both must be fixed before merge.

blackhathckr (approved):
## R2 Re-Review: Engagement Dashboard

All R1 issues fixed by Ananya. R2 issues (missing RPC migration + raw streak query) fixed directly.

- Migration added: `20260510_engagement_dashboard_rpcs.sql` with all 6 RPCs
- Streak computation moved server-side via `get_streak_distribution` RPC
- No remaining blockers, majors, or minors. Approving.

PR #281, opened 2026-05-15
fix: admin league creation rest_days default and status constraint

ananyactrl:
## Summary

Fix admin league creation failing with two DB constraint violations.

## What was broken

1. `rest_days` defaulted to 18 in both the service layer and form dialog. The `leagues` table has `CHECK (rest_days >= 0 AND rest_days <= 7)` because `rest_days` is per-week, not total. Every admin league creation failed with: `leagues_rest_days_check`

2. `status` was hardcoded to `'draft'` in the service insert. The current schema constraint only allows `('scheduled','active','ended','completed','cancelled','abandoned')` — `'draft'` was removed in the v2.5 migration. Every admin league creation failed with: `leagues_status_check`

## What was fixed

- `admin-leagues.ts`: `rest_days` default 18 → 1, clamped to max 7; `status` `'draft'` → `'scheduled'`
- `league-form-dialog.tsx`: `rest_days` default 18 → 1 in both initial state and reset-on-open; added `max={7}` and `Math.min` clamp; label shows valid range (0-7)

## Tested

- Admin league creation via `/admin/leagues` — league creates successfully without constraint errors

## Impact

- Web admin only — no mobile or player-facing impact

## Risk

- Low — only changes defaults to match existing schema constraints

## Files Modified

- `src/lib/services/admin/admin-leagues.ts`
- `src/components/admin/league-form-dialog.tsx`


blackhathckr (changes requested):
Two issues to fix:

1. **No lower-bound clamp on `rest_days`** — `Math.min(value, 7)` clamps the max but not the min. A negative value (e.g., via API/devtools) bypasses `min={0}` HTML attribute and hits the DB constraint `CHECK (rest_days >= 0 AND rest_days <= 7)`. Use `Math.max(0, Math.min(value, 7))` in the service layer.

2. **Update path unprotected** — `mapAdminLeagueInputToDbUpdates` passes `rest_days` directly to Supabase without clamping. An admin editing a league's rest_days could hit the same constraint error. Add the same clamping in the update path.

blackhathckr (changes requested):
Good fix for the DB constraint violations. Two items from the May 17 review are still unaddressed:

1. **No lower-bound clamp** — `Math.min(value, 7)` prevents values above 7 but doesn't prevent negatives. A negative value submitted via API/devtools bypasses the HTML `min={0}` attribute and hits the DB `CHECK (rest_days >= 0)` constraint. Fix:
   ```ts
   Math.max(0, Math.min(value, 7))
   ```
   Apply this in both the input handler and the data loading path.

2. **Update path unprotected** — `mapAdminLeagueInputToDbUpdates` passes `rest_days` directly to Supabase without clamping. Apply the same `Math.max(0, Math.min(value, 7))` there.

Both are quick fixes — please push.

blackhathckr (approved):
Both items addressed:

1. Lower-bound clamp added via clampRestDays(value) = Math.max(0, Math.min(value, 7)). Used in createLeague.
2. Update path now clamps in mapAdminLeagueInputToDbUpdates.

The service layer provides the real guard — HTML min={0} and form handler cover the UI side. This is fine.

Approved.

PR #282, opened 2026-05-15
fix: team invite validation crash when team in multiple leagues

ananyactrl:
## Summary

Fix team invite validation crashing when a team exists in multiple leagues.

## What was broken

`validateTeamInviteCode` used `.single()` on the `teamleagues` join query. If a team existed in more than one league, Supabase returned multiple rows and threw a `PGRST116` error, making all team invite links return invalid.

## What was fixed

Changed to `.order('created_at', { ascending: false }).limit(1).maybeSingle()` — always picks the most recent league association without crashing on multi-league teams.

## Tested

- Team invite validation and join via mobile and web — invite links work correctly

## Impact

- Mobile and web team invite flows

## Risk

- Low — `.maybeSingle()` returns `null` for no-match instead of throwing, which is handled by the existing null check

## Files Modified

- `src/lib/services/team-invites.ts`


blackhathckr (approved):
LGTM. Clean fix — .single() replaced with .order().limit(1).maybeSingle() correctly handles teams in multiple leagues. Consistent with existing patterns in the codebase.

PR #283, opened 2026-05-15
fix: supabase image domains and league create no-tier UX

ananyactrl:
## Summary

Fix Supabase image blocking, league create no-tier UX, and submissions table React warning.

## What was broken

1. Images served from Supabase storage were blocked by Next.js — `remotePatterns` for `*.supabase.co` was missing from `next.config.mjs`
2. League create page showed a blank summary card when no pricing tiers were configured, with no explanation to the user
3. Submissions table had a column missing an `id` key, causing a React warning

## What was fixed

- `next.config.mjs`: Added Supabase storage remote patterns
- `src/app/(app)/leagues/create/page.tsx`: Show summary card with clear warning when no pricing tiers exist
- `src/app/(app)/leagues/[id]/submissions/page.tsx`: Added `id: 'date'` to submissions table column definition

## Tested

- League create page with and without pricing tiers — warning card shows when no tiers configured
- Submissions table renders without React warnings
- Supabase images load correctly

## Impact

- Web only — no mobile impact

## Risk

- Low — image patterns are additive, UI changes are conditional

## Files Modified

- `next.config.mjs`
- `src/app/(app)/leagues/create/page.tsx`
- `src/app/(app)/leagues/[id]/submissions/page.tsx`


blackhathckr (changes requested):
Major issue — dead-end UX when no tiers exist:

The "No pricing tiers configured. League will be created for free." message renders, but the submit button stays permanently disabled because `pricePreview` is `null` when no tier is selected, and `handleCreateLeague` early-returns with an error if `!selectedTierId`.

User sees "will be created for free" but can't actually create. Either:
- Enable the button and handle no-tier creation in `handleCreateLeague` (create a free league), or
- Change messaging to "League creation unavailable until tiers are configured" and hide the button

The image config and submissions `id` fix are fine.

blackhathckr (changes requested):
The `handleCreateLeague` logic changes are correct, but **the original dead-end UX issue is not fully resolved** — the submit button is still disabled when no tiers exist.

### Original issue — still broken

The submit button `disabled` prop includes `!validation?.valid`. When `tiers.length === 0`, the price preview effect returns early and `validation` stays `null`. Since `null?.valid` is `undefined` and `!undefined` is `true`, **the button remains disabled**. The user sees "No pricing tiers configured. League will be created for free." but cannot click the button.

**Fix** — update the `disabled` condition to skip validation when no tiers exist:
```jsx
disabled={loading || (tiers.length > 0 && !pricePreview) || (tiers.length > 0 && !validation?.valid)}
```

Or alternatively, set `validation` to `{ valid: true, errors: [], warnings: [] }` in a useEffect when `tiers.length === 0`.

### Other items

- ✅ `handleCreateLeague` early return — fixed (`tiers.length > 0 && !selectedTierId`)
- ✅ `tier_id` omitted when no tier — fixed (spread conditional)
- ✅ Free-path routing — fixed (`!selectedTierId || ...`)
- ✅ Button label — correctly shows "Create League" vs "Pay & Create"
- ⚠️ Footer text still says "Payment required to create" even on the free path — misleading
- ⚠️ `next.config.mjs` has both a specific Supabase hostname and `*.supabase.co` wildcard — the specific one is redundant

blackhathckr (changes requested):
The handleCreateLeague logic fix from round 1 is good, but the core dead-end is still present:

1. **BLOCKER — Submit button permanently disabled in no-tier scenario.** The button's `disabled` prop includes `!validation?.valid`. When `tiers.length === 0`, the price preview effect returns early, so `validation` stays `null` → `!null?.valid` → `!undefined` → `true` → button disabled forever. Fix:
   ```jsx
   disabled={loading || (tiers.length > 0 && !pricePreview) || (tiers.length > 0 && !validation?.valid)}
   ```

2. **Minor — Footer text misleading.** Footer still says "Payment required to create" on the free path. Should be conditional.

3. **Minor — Redundant image pattern.** The specific hostname `umxfgpwvsgorkwytgbcc.supabase.co` is redundant since `*.supabase.co` already covers it. Remove the specific entry.

Item 1 is the blocker — the "create free league" path is completely unusable. Please fix.

blackhathckr (approved):
All three items from the previous review are fixed:

1. Submit button is no longer permanently disabled in no-tier scenario — the disabled logic now correctly gates on tiers.length > 0.
2. Footer shows "Free tier — no payment required" when total is 0 or no tiers exist.
3. Redundant specific Supabase hostname removed; only the wildcard remains.

Minor cleanup: there are two images: keys in the nextConfig object in next.config.mjs — the second one silently overwrites the first. Remove the first images block entirely to avoid confusion. Not blocking.

Approved.

PR #284, opened 2026-05-15
fix: incomplete profile banner not disappearing even after completing profile

dhjr:
## Summary
This PR fixes a critical regression where the Profile Completion Banner was failing to appear for incomplete profiles. It also addresses the root cause of previous "false positive" alerts by ensuring the banner logic is perfectly synchronized with the user's session data and the global application state.

## Core Issues Fixed
- **False Negatives (Not Showing):** Users with incomplete profiles were not seeing the banner because:
  - Auto-redirects from the Dashboard were moving users away from the only place the banner was rendered.
  - Dismissal state was shared across all users on the same browser, causing the banner to stay hidden for new test profiles if it had been dismissed once.
- **False Positives (Showing when Complete):** Previously, the banner incorrectly flagged users as incomplete (particularly for Date of Birth) because of mismatches between the database and the client-side session.

## Key Changes
- **Global Deployment:** Moved the banner to `AppLayoutClient` so it remains visible across the entire application until the profile is 100% complete.
- **Synchronized Validation:** Re-aligned the `PROFILE_FIELDS` check functions to precisely match the fields exposed in the NextAuth session (`name`, `phone`, `date_of_birth`, `profile_picture_url`).
- **User-Specific Dismissal:** Scoped the `localStorage` dismissal key to the User ID (`mfl_profile_banner_dismissed_[userId]`). This prevents one user's dismissal from affecting others and ensures test profiles see the banner as intended.
- **Hydration Safety:** Refactored the component to use a mounted state check. This ensures that it only renders after the client-side session is fully loaded and validated, preventing incorrect "incomplete" states during page transitions.


blackhathckr (approved):
LGTM. Banner correctly moved to global layout, date_of_birth threaded through auth pipeline, user-scoped localStorage key prevents cross-user contamination, and hydration safety is handled properly.

PR #285, opened 2026-05-15
fix: add start date validation and handling for empty or invalid date states in quick start wizard

dhjr:
## 🎯 Objective
Resolve a critical UI crash in the **Quick Start Wizard** Step 2 (Basics) that occurs when a user clears the start date field.
## 🐛 The Problem
When using the native browser date picker, clicking the "Clear" (x) button sets the input value to an empty string (`""`). The `QuickStartWizard` components were attempting to:
1.  Construct a `new Date()` from the empty string.
2.  Pass that `Invalid Date` into the `date-fns` `format()` function.
3.  `date-fns` throws a `RangeError` on invalid dates, which triggered the global Next.js error boundary ("Something went wrong").
### 1. Guarded Date Calculations
Added null/validity checks in `useMemo` hooks before attempting to format dates in:
- `StepBasics.tsx`
- `StepReviewLaunch.tsx`

### 2. UI Fallbacks
Implemented conditional rendering to handle missing dates gracefully:
- **Step 2 (Basics):** Shows a red instruction message ("Please pick a valid start date") instead of the "Ends on" calculation when the start date is empty.
- **Step 5 (Review):** Displays "Missing" in red for both start and end dates if they are unavailable.
### 3. Navigation & Submission Validation
Added explicit checks to prevent moving forward or submitting without a valid date:
- Added a validation toast in `StepBasics.tsx`'s `handleNext`.
- Added a final safety check in `QuickStartWizard.tsx`'s `handleSubmit`.


blackhathckr (approved):
LGTM - clean, minimal fix for the date-picker crash. Guards are correct, validation prevents forward navigation without a valid date, no regressions.

blackhathckr (approved):
LGTM. Clean fix for date picker crash — guards are solid, defense-in-depth with step-level and submit-level validation.

PR #287, opened 2026-05-15
Signup/form validation

dhjr:
# PR Description: Auth Validation, League Creation Improvements & UX Fixes
This PR addresses several UX and validation issues across the authentication flow, league creation wizard, and mobile layout.
### 1. Authentication & Form Validation
- **Email Validation**: Implemented regex-based email validation in both `SignupForm` and `LoginForm`.
- **Phone Number Restrictions**:
  - Limited input to **10 digits** only.
  - Automatically filters out non-numeric characters during input.
- **Username Restrictions**:
  - Limited usernames to **letters, numbers, and underscores** only.
  - Added a real-time character filter and a descriptive hint below the field.
- **Smart Submit Buttons**: 
  - The "Continue" (Signup) and "Login" buttons are now **disabled** until all mandatory fields are filled.
### 2. League Creation Improvements
- **Mandatory League Name**: 
  - Changed the League Name from optional to **mandatory** in both the standard creation flow and the **Quick Start Wizard**.
  - Removed automatic fallback names to ensure users explicitly name their leagues.
  - Added visual `*` indicators to required fields.
### 3. PWA Installation dialog box & minor related fixes
- **Mobile Layout Fix**: Resolved a horizontal overflow issue on mobile caused by the `PwaInstallPrompt`.
- **Nesting**: Moved the prompt inside the `SidebarInset` to ensure correct layout flow.
- **Padding**: Added `pb-20` on mobile to prevent the prompt from overlapping with the bottom navigation tabs.
### 4. Image optimization and domain whitelisting
- **Image Whitelisting**: Updated `next.config.mjs` to whitelist the Supabase storage domain (`umxfgpwvsgorkwytgbcc.supabase.co`).
- **Optimization**: This allows the app to use the optimized `next/image` component for profile pictures instead of standard fallback tags.
### Automated Tests
- Validated email regex against multiple valid/invalid formats.
- Verified character filtering logic for phone and username fields.


blackhathckr (changes requested):
Three items to address:

1. **Username max-length** — No `maxLength` on the username input or length check in `validateForm`. If the DB column has a limit (e.g., varchar(50)), long usernames will fail with an unclear error. Add a `maxLength` attribute and/or validation.

2. **Dead `autoName` code** — In `StepBasics.tsx`, the `autoName` memo is still computed but only drives the placeholder. Either remove it entirely or make it explicit that it's placeholder-only.

3. **PR description mismatch** — The description mentions PWA/layout changes and `PwaInstallPrompt` that are not present in this diff. Please trim the description to match actual changes.

blackhathckr (changes requested):
The two substantive code issues are properly fixed. Minor items remain.

### Original issues

1. ✅ **Username max-length** — Fixed comprehensively: `USERNAME_MAX_LENGTH = 30` constant, `maxLength` attribute on input, length range validation in `validateForm()`, regex validation (`/^[a-zA-Z0-9_]+$/`), and real-time stripping of invalid characters in `onChange`. Good work.

2. ✅ **Dead `autoName` code** — Removed entirely from `StepBasics.tsx`. League name is now mandatory with proper validation. `QuickStartWizard.tsx` updated to throw if `effectiveName` is empty.

3. ❌ **PR description mismatch** — Still references "PWA & Layout Fixes", `PwaInstallPrompt`, `SidebarInset`, and `pb-20` changes that do not exist in the diff. Please trim sections 3 and 4 from the PR description to match the actual changes.

### Notes

- ⚠️ `PHONE_LENGTH = 10` is hardcoded — this is fine if the product is India-only, but worth adding a comment noting the assumption. International numbers would be rejected.
- ⚠️ Minor inconsistency: `StepBasics` validates with `toast.error()` while `QuickStartWizard` uses `throw new Error()` in a try/catch. Both work but the pattern is inconsistent.

blackhathckr (changes requested):
Code fixes from round 1 are good — username maxLength and dead autoName code are resolved. One remaining item:

1. **PR description mismatch** — The body still references "PWA & Layout Fixes" (section 3) and parts of "Infrastructure & Performance" (section 4) that are not in the diff. Please update the PR description to only cover what's actually in this PR:
   - Signup form validation (email, username, phone)
   - Login form validation
   - League name made mandatory in QuickStart
   - Supabase image domain whitelisting

This is a documentation-only fix — just edit the PR body. No code changes needed.

blackhathckr (changes requested):
## Review: REQUEST CHANGES

Good direction — real-time input sanitization, `maxLength` attributes, and making league name required are all solid improvements. One blocking issue:

### Blocking Issue:

**No server-side validation mirrors the client-side constraints (Security)**
The PR adds client-side validation for email format (`EMAIL_REGEX`), username format (alphanumeric + underscore, 3-30 chars), and phone format (10 digits). But `src/app/api/auth/signup/route.ts` and `src/lib/services/auth-service.ts` only check for empty strings — no format/length validation.

Client-side validation is trivially bypassed with a direct API call. A malicious user could submit any string as username and it would be stored in the DB. While React escapes by default and Supabase parameterizes queries, server-side validation is still required as defense-in-depth.

**Fix:** Mirror the validation constants (`EMAIL_REGEX`, `USERNAME_REGEX`, `USERNAME_MIN_LENGTH`, `USERNAME_MAX_LENGTH`, `PHONE_LENGTH`) on the server side in the signup route.

### Minor Issues:

1. **`PHONE_LENGTH = 10` is India-specific** — if intentional, add a comment. If not, consider supporting country codes.
2. **`isFormComplete` in login-form** checks `!!password` but not trimmed — a spaces-only password passes the check (functionally harmless since auth would fail, but inconsistent).
3. **`isDetailsComplete` doesn't check partial phone input** — user can type 3 digits and button enables, then validation blocks on submit. Consider disabling when phone has 1-9 digits.

### What looks good:
- `onChange` handlers strip disallowed chars in real-time (username, phone) — excellent UX
- `maxLength` on inputs — good defense-in-depth
- `FieldDescription` hint below username field — helpful
- Duplicate `images` block removed from `next.config.mjs` — real bug fix
- League name now required in quick-start wizard — good product decision

blackhathckr (changes requested):
## Re-review: REQUEST CHANGES

Server-side validation added to signup route — good. Shared constants (`EMAIL_REGEX`, `USERNAME_REGEX`, etc.) properly imported and checked. However:

### Still needed:

**1. Missing password length check in signup route**
Client-side checks `MIN_PASSWORD_LENGTH` but server accepts any non-empty password. Add:
```ts
if (password.length < MIN_PASSWORD_LENGTH) {
  return NextResponse.json({ error: 'Password must be at least...' }, { status: 400 });
}
```

**2. `/api/auth/verify-otp` has zero field validation**
This is an alternative user-creation endpoint (OTP flow with `createUser: true`). It accepts username, password, phone, email — with no format/length checks. Same validation constants should be applied here too, otherwise the signup validation is bypassable.

**3. `/api/auth/complete-profile` lacks username/phone validation**
Accepts username and phone without format checks.

The signup route fix is solid — just extend the same pattern to the other auth endpoints that accept the same fields.

blackhathckr (approved):
## Re-review: APPROVE ✅

All previously flagged issues resolved:

1. ✅ **Password length check** — `MIN_PASSWORD_LENGTH` now enforced in signup route
2. ✅ **`/api/auth/verify-otp`** — Full validation added (email, username, password length, phone)
3. ✅ **`/api/auth/complete-profile`** — Same validation applied (username, password, phone)
4. ✅ Shared constants (`EMAIL_REGEX`, `USERNAME_REGEX`, `PHONE_LENGTH`, `MIN_PASSWORD_LENGTH`) used consistently across all auth endpoints

Server-side validation now mirrors client-side constraints across all user-creation paths. Good to merge.

PR #290, opened 2026-05-16
feat: enforce exclusive submission rules between workouts and rest days

dhjr:
## Description
This PR implements frontend-side mutual exclusivity between activity submissions and rest days. It prevents users from logging both a workout and a rest day on the same date, providing a better UX than waiting for a backend failure.

---

## Key Changes

*   **Exclusivity Logic:** Added `hasRestDayOnDate` and `hasWorkoutOnDate` memos to track existing submissions for the selected date.
*   **Dynamic Disabling:**
    *   The "Activity" tab is disabled if a rest day exists for the selected date.
    *   The "Rest Day" tab is disabled if one or more workouts exist for the selected date.
*   **UI Alerts:** Added informational alert banners in both tabs to explain why a submission type is unavailable.
*   **Form Overlay:** Applied an opacity and pointer-events overlay to the form when it is blocked by a conflicting submission.
*   **Rejection Handling:** Ensured that `rejected` or `rejected_resubmit` entries do not trigger exclusivity, allowing users to switch submission types if a previous attempt was unsuccessful.


blackhathckr (changes requested):
Critical regression:

**`suspiciousProofStrikes` → `suspicious_proof_strikes` will break the suspicious proof warning feature.**

The API at `/api/leagues/[id]/my-submissions/route.ts` returns this field as camelCase (`suspiciousProofStrikes`). The other two fields on lines 165-169 (`suspiciousProofWarningThreshold`, `suspiciousProofRejectionThreshold`) remain camelCase, confirming this is a mistake. After merge, `suspicious_proof_strikes` will always be `0` via the `?? 0` fallback, silently disabling warnings.

Fix: Revert to `json.data.suspiciousProofStrikes`.

Also: Remove unused `import Image from 'next/image'` in `dashboard-header.tsx`.

The exclusivity logic itself is solid — good defense-in-depth with both frontend and backend enforcement.

blackhathckr (changes requested):
Original issues are fixed, but a new critical regression was introduced.

### Original issues

1. ✅ **`suspiciousProofStrikes` property name** — Correctly reverted to camelCase (`json.data.suspiciousProofStrikes`), matching the API response format.
2. ✅ **Unused `Image` import** — Removed from `dashboard-header.tsx`.

### New critical issue — auto-assigned rest day override is broken

The new exclusivity check at ~line 760 in `src/app/api/entries/upsert/route.ts` returns a 409 when `type === 'workout' && existingRestEntry`. This fires **before** the existing auto-rest-day cleanup logic at ~line 1044, which is specifically designed to delete cron-assigned rest days and allow late workout submissions.

**Result:** Users who missed a day (and had a rest day auto-assigned by the cron job) can no longer submit a late workout for that date. The auto-rest cleanup code is now unreachable for this scenario.

**Fix** — exclude auto-assigned rest days from the exclusivity check:
```javascript
const existingRestEntry = nonRejectedEntries.find(
  (r) => r.type === 'rest' && !(r.notes?.includes('Auto-assigned'))
);
```

Or alternatively, move the exclusivity check **after** the auto-rest cleanup block.

### Minor issues

- ⚠️ `pointer-events-none` on the form overlay prevents mouse clicks but not keyboard interaction. The submit buttons don't include `hasRestDayOnDate`/`hasWorkoutOnDate` in their `disabled` prop — a user could Tab + Enter to submit. Backend catches it (409), but the UX is inconsistent. Add the exclusivity flags to the button `disabled` conditions.

blackhathckr (changes requested):
Round 2 feedback still outstanding. The exclusivity logic is solid but there's a critical regression:

1. **CRITICAL — Auto-assigned rest day override is broken.** Your 409 check at ~line 760 (`if (type === 'workout' && existingRestEntry)`) fires BEFORE the existing auto-rest-day cleanup logic at ~line 1044. Users who had a rest day auto-assigned by the cron job can no longer submit a late workout — the 409 blocks them. Fix: either exclude auto-assigned rest days from the exclusivity check (e.g., filter out entries where `notes?.includes('Auto-assigned')` or better, check a dedicated flag), or move the exclusivity check after the auto-rest cleanup block.

2. **Submit buttons not disabled** — The workout submit button's `disabled` prop doesn't include `hasRestDayOnDate`, and vice versa. A user can Tab+Enter past the `pointer-events-none` overlay. The backend 409 catches it but the UX is confusing. Add the flags to the `disabled` prop.

The auto-rest regression is a blocker — it breaks a legitimate user workflow. Please fix before this can be merged.

blackhathckr (approved):
Both items fixed:

1. Auto-assigned rest day override regression — fixed. Backend now detects auto-assigned rest days via notes.includes('Auto-assigned') and allows workouts to override them while still blocking manual rest days. Good approach.

2. Keyboard bypass — fixed. Submit buttons now include hasRestDayOnDate / hasWorkoutOnDate in the disabled prop directly, so Tab+Enter can't bypass the overlay.

Defense-in-depth is solid: tabs disabled + form overlay + submit button disabled + backend 409.

Approved.

PR #293, opened 2026-05-17
Fix/UI cleanup and navigation

dhjr:
## Summary
This PR addresses several UI, navigation, and user flow improvements. It resolves layout issues on the leaderboard and mobile tabs, adds a missing navigation link to the Analytics page for authorized league managers, and removes auto-redirection from the main dashboard.

## Changes Made
- **Leaderboard Cleanup:** Removed the duplicate static header from `src/app/(app)/leagues/[id]/leaderboard/page.tsx` and updated the `Suspense` fallback to display the proper skeleton component while loading.
- **Mobile Bottom Nav Fix:** Addressed horizontal scrolling issues and off-screen rendering in `MobileBottomTabs` by strictly enforcing `w-full`, adding `min-w-0` to tab items to allow proper flex shrinking, and replacing a custom class with the standard `pb-[env(safe-area-inset-bottom)]` for iOS home indicator support. 
- **Analytics Access:** Added an "Analytics" link to the `oversightItems` block in `src/lib/navigation/sidebar-config.ts` so Hosts and Governors can easily access the analytics dashboard directly from the main sidebar.
- **Dashboard Auto-Redirection Removed:** Removed the logic in `src/app/(app)/dashboard/page.tsx` that previously forced users with only a single active league to skip the MFL dashboard. Now all users can view their MFL dashboard, regardless of how many leagues they are in.

## How to Test
1. **Analytics:** Log in as a Governor or Host and verify that the "Analytics" button appears in the sidebar under "Oversight Actions" and successfully routes to `/leagues/[id]/analytics`.
2. **Leaderboard:** Navigate to the Leaderboard on desktop and verify that only the single, styled header (with filters) is rendered.
3. **Mobile Layout:** Open the app on a mobile view (using DevTools or a real device) and verify there is no horizontal scrolling across different pages, and that the bottom navigation bar is perfectly anchored to the bottom.
4. **Dashboard Redirect:** Log in as a user with exactly one active league. Verify that you land on the `/dashboard` page and are not immediately redirected to the league view.


blackhathckr (changes requested):
First review. The individual changes are clean — leaderboard dedup, mobile bottom nav fix, analytics sidebar link, mandatory league name are all good. A few items:

1. **Verify auto-role-switch side effect** (`use-league-settings-form.ts`): The new `useEffect` silently switches the user to `host` role when they access settings. If a player navigates to settings, their role changes globally. Confirm this is intentional and doesn't affect other parts of the UI that depend on `currentRole`. If it should only apply within settings, consider a local role override instead of mutating global state.

2. **Price display change** (`tier-recommendation-card.tsx`): You switched from `tier.pricing.fixed_price` to `priceBreakdown.subtotal`. Verify that `priceBreakdown.subtotal` is always present when `pricing_type === 'fixed'` — if it can be `null`/`undefined`, the price will render blank.

3. **Confirm `Activity` icon import** (`sidebar-config.ts`): The diff shows `Activity` used as the analytics icon but doesn't show the import statement. Make sure it's imported from `lucide-react` at the top of the file.

4. **Minor — En-dash removal** (`residential-pricing.tsx`): You changed `"40–150 Residents"` to `"40-150 Residents"`. En-dashes are typographically correct for ranges. Consider keeping the en-dash unless there's a specific reason to remove it.

Items 1-3 need confirmation/verification. No hard blockers but please respond.

blackhathckr (approved):
## Review: APPROVE ✅

Clean, well-scoped PR. 10 files, +107/-100 lines. All changes are reasonable UI/UX improvements.

### What looks good:
- Dashboard single-league redirect removal — clean
- Leaderboard duplicate header removal — correct
- Tier recommendation card refactor — price display fix using `priceBreakdown.subtotal` is more correct
- Mobile bottom tabs CSS fixes — `pb-[env(safe-area-inset-bottom)]` is proper
- Custom activities routing change — consistent with new page structure
- Analytics nav item addition — icon already imported, target page exists

### Minor items (non-blocking, can fix post-merge):
1. **`page-old.tsx`** in `leagues/create/` will have TS errors since you removed `isLoading`/`onChangeTier` props from `TierRecommendationCard`. Consider deleting that dead file.
2. **Auto role-switch** in `use-league-settings-form.ts` — the `useEffect` that auto-switches to `host` role when a host user accesses settings could surprise users who intentionally selected a different role. Worth a follow-up discussion on whether a local override is better.

No security issues, no debug artifacts.

blackhathckr (changes requested):
## ⚠️ Updated Review: REQUEST CHANGES (was APPROVE)

Deeper analysis found a **blocking UX regression**:

### Blocking Issue:

**Removing "Confirm Recommendation" button breaks league creation**
In `leagues/create/page.tsx`, the "Confirm Recommendation" button was the primary way to set `selectedTierId` when a user accepts the recommended tier. With it removed, the only way to set `selectedTierId` is through the TiersModal. But the validation at ~line 382 (`if (tiers.length > 0 && !selectedTierId)`) requires it to be set — so users who accept the recommendation without opening "View All Tiers" will get a "Please select a tier" error and can't create the league.

**Fix:** Either auto-set `selectedTierId` to the recommendation's `tier_id` when the recommendation loads, or restore the confirm button.

### Additional Issues Found:

1. **Dead import in `settings-section-page.tsx`** — `CustomActivitiesSection` still imported but its case was removed from the switch statement.
2. **Nested `@container/main`** — `LeaderboardPageSkeleton` wraps in `@container/main` but is already inside a parent `@container/main` div in the leaderboard page.
3. **`page-old.tsx` will fail TS** — still passes `isLoading`/`onChangeTier` props that were removed from `TierRecommendationCard` interface. Delete or update.

Previous review's notes on auto role-switch and other items still apply.

blackhathckr (approved):
## Re-review: APPROVE ✅

All blocking issues resolved:

1. **Tier auto-selection** — `selectedTierId` now auto-sets to `recommendation.tier_id` via useEffect. No more broken happy path. Clean fix.
2. **Dead import** — `CustomActivitiesSection` import and switch case removed from `settings-section-page.tsx`.
3. **`page-old.tsx`** — Gutted to a 3-line stub (returns null). Ideally delete entirely in a follow-up, but not blocking.
4. **`@container/main` nesting** — Fixed in `LeaderboardPageSkeleton`.

Minor nits (non-blocking, follow-up):
- Delete `page-old.tsx` stub entirely
- Narrow recommendation effect deps from `formData` object to specific fields (`formData.num_teams`, `formData.max_participants`) to avoid double-firing

Good to merge.

PR #294, opened 2026-05-17
Fix/challenges rendering issue

dhjr:
## Summary
This PR resolves a crashing bug in the AI Manager tab and fixes a critical issue where newly created custom challenges were failing to appear on the Configure Challenges page for league admins.

## Description
- **AI Manager Crash Fix:** Fixed a `TypeError` (`activeChallenges.filter is not a function`) in `src/components/ai-manager/challenges-tab.tsx`. The component was incorrectly attempting to fall back to an empty object instead of an array when parsing the API response. Added an explicit `Array.isArray()` check to guarantee state integrity.
- **Challenge Creation Visibility Fixes:** Resolved multiple bugs that prevented newly created challenges from rendering on the frontend:
  1. **Role Hierarchy Bug:** Fixed a flaw in `getMembership` (`api/leagues/[id]/challenges/route.ts` and `[challengeId]/route.ts`). Previously, if a user was both a Host and a Player, the API might randomly evaluate them as just a "Player" due to unordered SQL joins, immediately stripping their admin privileges and hiding their newly created `draft` challenges. The backend now strictly sorts roles by privilege hierarchy (`host > governor > captain > player`) to guarantee users retain their highest permissions.
  2. **Schema Mismatch:** Removed a reference to an `is_unique_workout` column from the `GET /challenges` API query. The column does not exist in the current Supabase schema, which was causing the database query to fail silently and return an empty `[]` array to the frontend.
  3. **Stale Cache Prevention:** Added `cache: 'no-store'` to the client-side `fetch` call in `src/app/(app)/leagues/[id]/configure-challenges/page.tsx` to prevent the browser from improperly caching empty lists after a challenge is created.

## How to Test
1. **AI Manager Stability:** Navigate to the AI Manager page in a league. Verify that the "Challenges" tab loads successfully without crashing, even if there are no active challenges.
2. **Challenge Creation:**
   - Log in as a user who has multiple roles (e.g., both a `Host` and a `Player` in the same league).
   - Navigate to the "Configure Challenges" page.
   - Click "Create Challenge" and fill out the details for a new custom challenge.
   - Submit the form. Verify that the newly created challenge appears instantly in the UI with a `Draft` status badge.
3. **Challenge Management:** Verify that as a Host/Governor, you can successfully click "Edit" or "Delete" on the newly created challenge without receiving a 403 Forbidden error from the backend.


blackhathckr (changes requested):
First review of this PR. The role priority fix and schema mismatch removal are good, but two blockers:

1. **BLOCKER — Debug code left in production.** `src/app/api/leagues/[id]/challenges/route.ts` contains:
   ```js
   require('fs').writeFileSync('/mnt/F65CB2605CB21AF7/MFL/mfl/debug.txt', ...)
   ```
   This will crash on Vercel (no filesystem write access) and must be removed immediately.

2. **BLOCKER — Challenges tab will show empty list forever.** Your fix checks `Array.isArray(data.data)` but the GET API returns `{ success: true, data: { active: [...], availablePresets: [...] } }`. `data.data` is an object, not an array, so `Array.isArray(data.data)` is `false`. Then `Array.isArray(data)` is also `false`. Result: `challenges` stays `[]` and the tab renders empty. The crash is fixed but the feature is non-functional. Fix should be something like:
   ```ts
   let challenges = [];
   if (data?.data?.active && Array.isArray(data.data.active)) {
     challenges = data.data.active;
   } else if (Array.isArray(data?.data)) {
     challenges = data.data;
   } else if (Array.isArray(data)) {
     challenges = data;
   }
   ```

3. **Minor — Duplicated `rolePriority` map.** The same priority map and sorting logic is copy-pasted in both `route.ts` and `[challengeId]/route.ts`. Extract to a shared utility (e.g., `src/lib/utils/roles.ts`).

4. **Minor — Diff noise.** ~3000 lines of the diff are pure re-indentation in `configure-challenges/page.tsx`. In future, please separate formatting changes from functional fixes — it makes review much harder.

Items 1 and 2 are blockers. Please fix.

blackhathckr (changes requested):
## Review: REQUEST CHANGES

The writeFileSync debug code from the previous review is gone — good. One remaining blocking bug:

### Blocking Issue:

**`is_unique_workout` removed from Supabase `.select()` but still used in response mapping**
In `src/app/api/leagues/[id]/challenges/route.ts`, the `.select()` column list no longer includes `is_unique_workout`, but the response mapping at ~line 285 still references `(c as any).is_unique_workout`. Since Supabase only returns selected columns, this will always be `undefined` → `false`. All "Unique Workout Day" challenges will silently appear as regular challenges, and the frontend checkbox will be wrong when editing.

**Fix:** Either add `is_unique_workout` back to the `.select()` string, or remove it from the mapping if intentionally deprecating the feature.

### What looks good:
- `rolePriority` sort fix — picks highest-priority role instead of arbitrary `roles[0]`
- Centralized `LeagueRole` type in `src/lib/utils/roles.ts` with `vice_captain`
- `challenges-tab.tsx` response parsing — more robust `data.data.active` handling with fallbacks
- `cache: 'no-store'` on frontend fetch
- No debug artifacts (no `console.log`, `writeFileSync`, etc.)
- All API routes properly authenticate and authorize

blackhathckr (approved):
## Re-review: APPROVE ✅

All previously raised issues are resolved:

1. **`is_unique_workout`** — back in the `.select()` query, consistent with the mapping. Fixed.
2. **`getMembership` centralized** — New shared utility `src/lib/utils/roles.ts` exports `getMembership()` with `rolePriority` sort. All 11 route files now import from it instead of duplicating the function. Excellent work.
3. **`challenges-tab.tsx`** — Robust response parsing with `data.data.active` and fallbacks.

Clean PR. Good to merge.

PR #297, opened 2026-05-22
fix: league launch

dikshaa-012:
## Summary
League launch was failing because the backend was writing `launched` as the status, which is not allowed by the DB constraint. Fixed by mapping `launch_meet` phase to `scheduled`.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] UI/UX improvement
- [ ] Documentation

## Changes Made
- Updated `launch_meet` → `scheduled` in `league-phases.ts` phase-to-status mapper
- Updated `launchLeague` in `leagues.ts` to use the correct status value

## How to Test
1. Log in as a league host
2. Navigate to a league in `mobilisation` phase
3. Tap Launch League
4. Verify the league transitions successfully without "Launch Failed" error

## Screenshots / Recordings
WhatsApp Image 2026-05-22 at 5 44 54 PM

## Checklist


blackhathckr (changes requested):
Good catch — the DB CHECK constraint doesn't allow 'launched' as a status value, so the fix to use 'scheduled' is correct. But there's a regression:

**LEGACY_STATUS_TO_PHASE is not updated.** You changed PHASE_TO_LEGACY_STATUS so launch_meet -> 'scheduled', but the reverse map LEGACY_STATUS_TO_PHASE still only has:
```
launched: 'launch_meet'
```

It has no entry for 'scheduled'. So any code using normalizeLeaguePhase() on a league with status='scheduled' will fall through to the default 'mobilisation' — launched leagues will appear as if they're still in mobilisation phase.

Fix: add this entry to LEGACY_STATUS_TO_PHASE:
```ts
scheduled: 'launch_meet',
```

Keep the existing 'launched' entry for backward compatibility. This is a one-line fix but without it, the launch transition will look like it worked but the phase display will be wrong everywhere.

blackhathckr (approved):
Both items addressed:

1. LEGACY_STATUS_TO_PHASE now has `scheduled: 'launch_meet'` — normalizeLeaguePhase will correctly resolve launched leagues.
2. Backward compat preserved — `launched: 'launch_meet'` entry kept for older leagues.

Clean, minimal fix. Approved.

PR #298, opened 2026-05-22
fix: analytics empty sections

dikshaa-012:
## Summary
The analytics page showed empty sections and errors instead of real data. Fixed several mismatched database references that were causing the entire page to fail silently.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] UI/UX improvement
- [ ] Documentation

## Changes Made
- Fixed wrong table name `resdays` → `rest_day_donations` in rest day records query
- Fixed wrong column name on `rest_day_donations` table
- Fixed daily submissions query ordering by nonexistent `entry_date` column → `date`
- Fixed performer username lookup — now fetches usernames separately instead of relying on a nested join that was silently failing

## How to Test
1. Log in 
2. Navigate to the Analytics page for a league with at least one member submission
3. Verify top/bottom performers show player data
4. Verify daily submissions section shows data
5. Verify no 500 errors in the backend terminal

## Screenshots / Recordings
WhatsApp Image 2026-05-23 at 1 22 43 AM

## Checklist


blackhathckr (changes requested):
Good work fixing the core bugs — wrong table name (resdays -> rest_day_donations), wrong ORDER BY column (entry_date -> date), and broken user join. The error handling additions are also welcome. Two items to address:

1. **N+1 query problem in teamStats** — The Promise.all iterates over every team and runs 2 queries per team (members + submissions). For a league with 20 teams = 40 extra queries. Fix: fetch all league members and submissions in one query each, then group by team_id in JavaScript.

2. **Redundant role lookup** — You're fetching assignedrolesforleague with the nested join roles(role_name), then ALSO querying the roles table separately for the same role_ids. Either trust the join or don't use it — don't do both. If the nested join doesn't work (like the users(username) case), remove it from the select and keep only the manual lookup.

Minor: the switch from .in('league_member_id', memberIds) to .eq('league_id', leagueId) for rest day donations changes the semantics — it could include donations from members who left the league. Confirm this is intentional.

blackhathckr (approved):
Both items from the last review are addressed:

1. N+1 query fixed — team stats now group from the already-fetched allMembers and userSubmissions using Maps. No more Promise.all with per-team queries. This is correct and much more efficient.

2. Redundant role lookup removed — now only uses the nested join roles(role_name) from assignedrolesforleague. No separate roles table query.

Also good:
- rest_day_donations query correctly kept member-based filtering (.in league_member_id)
- Host exclusion from performers is clean
- Error handling is comprehensive throughout

Approved.

PR #302, opened 2026-05-23
fix: challenges render

dikshaa-012:
## Summary
Challenges were not appearing on the configure screen and could not be edited after creation. Two separate issues were causing this — a missing DB column being referenced in queries, and draft challenges being excluded from the host view.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] UI/UX improvement
- [ ] Documentation

## Changes Made
- Removed references to non-existent `is_unique_workout` column from GET and PATCH handlers in challenges route
- Updated GET `/api/leagues/[id]/challenges` to return a `draft` bucket alongside `active` for hosts/governors
- Updated configure challenges page to merge draft challenges into the displayed list

## How to Test
1. Log in as a league host
2. Navigate to Configure Challenges
3. Create a new challenge — verify it appears immediately in the list
4. Edit the challenge — verify it saves without error
5. Verify active challenges also still render correctly

## Screenshots / Recordings
WhatsApp Image 2026-05-24 at 2 16 13 AM

## Checklist


blackhathckr (approved):
## Review: APPROVE ✅

The `is_unique_workout` column doesn't exist in the DB schema — having it in the `.select()` was causing the challenges query to silently fail/return empty, which is why the challenges tab showed nothing. Correct to remove it from both GET select and PATCH handler.

Minor follow-up: the mapping at ~line 285 still references `(c as any).is_unique_workout` — it will always be `false` now, which is harmless. Clean it up in a follow-up if the feature is fully deprecated.

PR #306, opened 2026-05-24
feat: implement email delivery timeout and suppress OTP send errors to prevent enumeration

dhjr:
## Summary
This PR fixes an issue where the `send-otp` API route would hang and eventually time out when attempting to send an email to a fake or invalid domain. It wraps the email sending process in a 5-second timeout and silently ignores delivery errors, returning a success response to the client to prevent email enumeration.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] UI/UX improvement
- [ ] Documentation

## Changes Made
- **Wrapped `sendEmail` with a timeout**: Added a 5-second `Promise.race` timeout around the `sendEmail` call in `src/app/api/auth/send-otp/route.ts` to prevent the SMTP server from hanging on unreachable domains.
- **Silently caught delivery errors**: Modified the `catch` block to suppress 500 errors if the email fails to send or times out.
- **Consistent success response**: The API now always returns `{ ok: true }` regardless of delivery status, enhancing security by preventing attackers from enumerating valid/invalid email addresses.
- **Preserved OTP logic**: OTP generation, database storage, and rate limiting remain fully functional and securely in place.

## How to Test
1. Navigate to the "Forgot Password" or login flow on your local instance.
2. Enter a fake email address with an invalid or non-existent domain (e.g., `test@invalid-domain-12345.xyz`).
3. Click to send the verification code.
4. Verify that the frontend immediately shows a success state (or moves to the next step) without a long delay or a 500 server error.
5. Check the database (`email_otps` table) to verify the OTP was still generated and stored properly.
6. Verify that valid emails still successfully receive the OTP within the 5-second window.

## Checklist


blackhathckr (approved):
**APPROVED** ✅

Good security improvement.

**What's good:**
- Anti-enumeration pattern: always returning `{ ok: true }` regardless of email delivery prevents attackers from probing which emails exist in the system
- 5s timeout via `Promise.race` prevents hanging requests when SMTP is slow/down
- OTP is stored in DB before email attempt, so the record exists even if delivery fails
- Error is still logged server-side (`console.error`) so delivery failures are detectable in monitoring

**Note (non-blocking):**
- If SMTP is consistently broken, users will get `{ ok: true }` but never receive their OTP — they'll be stuck. Make sure there's server-side alerting on repeated `"Failed to send OTP email or timed out"` log entries.
- Most of the diff is formatting/line-wrapping — the actual functional change is small (timeout + silent catch). Clean.

PR #307, opened 2026-05-24
feat: implement 10-second fallback refetch timeout for realtime message updates

dhjr:
## Summary
This PR fixes an issue where sent chat messages would fail to appear immediately if the Supabase realtime connection silently stalled. It implements a 10-second heartbeat check that forces a manual refetch if the realtime confirmation event is missed, ensuring a seamless user experience without waiting for the 30-second global fallback polling.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [x] UI/UX improvement
- [ ] Documentation

## Changes Made
- **Added `onMessageSendStart` callback in `message-input.tsx`**: Exposes a hook to notify the parent container the exact moment a user initiates a message send.
- **Implemented Heartbeat Timer in `chat-window.tsx`**: Added a 10-second timeout that starts via `onMessageSendStart`. If the timer expires without a response, `fetchMessages(false)` is called to manually sync the chat state.
- **Added Event Cleanup**: Placed `clearRealtimeTimeout()` inside the Supabase `postgres_changes` listener. If the realtime event successfully arrives within the 10-second window, the timer is cleared to prevent unnecessary API requests.

## How to Test
1. Open the team chat in your local environment.
2. Send a message to verify the optimistic update appears instantly.
3. To test the stall, you can simulate a broken websocket connection (e.g., block the Supabase realtime domain in your browser's network tab *after* loading the page but *before* sending a message).
4. Send another message. Observe that exactly 10 seconds later, the chat manually refetches and the message is properly confirmed, avoiding the full 30-second fallback wait.
5. Restore the connection and send a message. Verify that the 10-second manual refetch *does not* occur (check network tab) because the realtime event cleared the timer.

## Checklist


blackhathckr (approved):
**APPROVED** ✅

Clean resilience pattern for realtime messaging.

**What's good:**
- 10-second fallback timer starts on message send, clears when realtime event arrives
- If Supabase realtime misses an event (which does happen), the UI still updates within 10 seconds via manual refetch
- Proper cleanup on unmount via `clearRealtimeTimeout` in useEffect
- Minimal footprint: `onMessageSendStart` callback threaded through cleanly to `MessageInput`
- No unnecessary re-renders — `useCallback` with correct dependencies

PR #310, opened 2026-05-24
fix: add missing is_unique_workout column to leagueschallenges table

ananyactrl:
## Summary

Fix challenge submission 404,  added missing `is_unique_workout` column to `leagueschallenges` table.

## What was broken

`POST /api/leagues/{id}/challenges/{challengeId}/submissions` returned 404 for all challenge submissions. The route's `ensureChallengeInLeague()` function selects `is_unique_workout` from `leagueschallenges`, but the column didn't exist in the database. Supabase returned an error, `data` came back null, and the route returned 404 "Challenge not found".

## What was fixed

Added migration to create `is_unique_workout` column with `DEFAULT false` so existing rows are unaffected. No application code changes needed — the route already handles the column correctly. 
Once applied, challenge submissions will work and can be tested on device.

## Tested

- Migration syntax validated
- Route code confirmed to handle `is_unique_workout` correctly

## Impact

- **Backend** — challenge submission endpoint

## Risk

- Low — column added with `DEFAULT false`, no existing data affected

## Files Modified

- `supabase/migrations/20260520_add_is_unique_workout_to_leagueschallenges.sql`


blackhathckr (approved):
**APPROVED** ✅

Adds the missing `is_unique_workout` column to `leagueschallenges` — this was always intended but the migration was never created, causing query failures.

- `DEFAULT false` means no impact on existing challenges
- `IF NOT EXISTS` makes it safe to re-run

**Follow-up needed:**
- PR #302 (Diksha) previously removed `is_unique_workout` from the challenges API route `.select()` and PATCH handler as a workaround for this missing column. Now that the column exists, it should be **added back** to the routes so the feature actually works end-to-end. Please raise a follow-up PR for that.

PR #320, opened 2026-05-24
fix: resolve OCR failures in serverless/production environment

ananyactrl:
## Summary

Fix OCR server failing on hosted/production — read-only filesystem crash, worker path resolution, and CJS compatibility.

## What was broken

OCR (`/api/ocr/extract`) worked locally but failed on the hosted `dev.myfitness` environment. Three root causes:

1. **Read-only filesystem crash** — `CACHE_DIR` used `process.cwd()/.cache/tesseract`. Vercel/Lambda filesystems are read-only except `/tmp`, so `mkdirSync` threw and killed the worker before initialization
2. **`eval('require')` unreliable** — Next.js CJS output doesn't guarantee `eval('require')` works. Worker script path resolution failed silently
3. **`import.meta.url` removed** — Not available in CJS-compiled Next.js server code, causing module resolution failures

## What was fixed

- `CACHE_DIR` now uses `/tmp/tesseract-cache` in production, falls back to cwd for local dev
- Replaced `eval('require')` with direct `path.join(process.cwd(), 'node_modules/tesseract.js/...')` — works in all compilation modes
- Added `../node_modules/...` as fallback for monorepo/hoisted installs
- Removed `import.meta.url` — cwd-relative approach is simpler and portable

## Note on pending workouts

The auto-approval behavior on the hosted dev environment is **not a bug** — the seed script assigns `host` role to test accounts on all leagues. The API correctly sees `host` → `isPrivileged = true` → auto-approved.

## Impact

- **Backend** — OCR API endpoint on production/hosted environments

## Risk

- Low — paths are environment-gated, local dev behavior unchanged
- `/tmp` is the standard writable directory for Vercel/Lambda

## Files Modified

- `src/lib/ocr/run-server-ocr.ts`


blackhathckr (approved):
LGTM — proper serverless fix. Using /tmp for cache dir in production, removing eval('require'), and worker candidate path resolution are all correct.

PR #322, opened 2026-05-25
fix: mark email_verified=true on mobile signup via OTP

ananyactrl:
## Summary

Fix mobile signup accounts stuck as unverified — pass `email_verified: true` when creating users via OTP verification.

## What was broken

Mobile signup via OTP created accounts with `email_verified = false` (DB default). No verification email was sent to mobile users, leaving accounts permanently stuck as unverified with no way to complete verification.

## Root Cause

The `verify-otp` route calls `createUserService` but never passes `email_verified: true`. Since OTP already proves email ownership, accounts created through this flow should be marked as verified immediately.

## What was fixed

- Added `email_verified?: boolean` to `createUser()` interface in `users.ts`
- Pass `email_verified: true` when calling `createUserService` in `verify-otp/route.ts`

## Related but separate issue

On `qa` deployment, the **web signup verification email** contains a `localhost:3000` link instead of the actual domain. This is caused by `siteConfig.url` not picking up the correct `NEXTAUTH_URL` env var on the qa deployment — needs a server-side env var update, not a code change. This is separate from the mobile OTP fix.

## Tested

- Mobile signup via OTP — account created with `email_verified = true`
- Existing accounts unaffected — migration not required, column default unchanged

## Impact

- **Mobile** — OTP signup flow (mobile only)
- Web signup verification email localhost issue is a deployment config issue, not a code fix

## Risk

- Low — OTP already proves email ownership, marking verified is the correct behavior
- No migration needed — column already exists

## Files Modified

- `src/lib/services/users.ts`
- `src/app/api/auth/verify-otp/route.ts`


blackhathckr (approved):
LGTM — OTP already validates email ownership, correct to set email_verified=true.

PR #327, opened 2026-05-26
fix(chat): resolve captains filter returning empty results due to con…

ananyactrl:
## Summary
Fix Captains filter in team chat returning empty results — messages sent as captain disappear after optimistic update and only appear under All after ~30s poll.

## What was broken
In `getMessagesForUser`, the Captains tab applied `visibility = captains_only` and then stacked a second `.or()` with more visibility conditions. In PostgREST, conflicting visibility filters cause the query to return no rows. The error was swallowed silently, returning an empty list. Host, Announcements, and Important filters worked because they don't add the second visibility `.or()`.

## What was fixed
When `filter === 'captains_only'`, scope by team and sender only — not visibility again. Also added error logging instead of failing silently.

## Tested
- Captains filter → captain sends message → stays in Captains tab immediately
- All other filters (All, Host, Announcements, Important) unaffected
- No mobile rebuild needed — API only

## Impact
Web/API — team chat message filtering

## Risk
Low — only changes query logic for captains filter, other filters untouched

## Files Modified
- `src/lib/services/messages.ts`


blackhathckr (approved):
LGTM — correct fix. The .eq('visibility', 'captains_only') was conflicting with the subsequent .or() visibility filter in PostgREST, returning empty rows. Separating the captains_only branch avoids the double-filter.

PR #416, opened 2026-06-12
fix: effective_points display + formula-gated scoring

blackhathckr:
## Summary
- Submission APIs return `null` for `effective_points` in standard/simple formula leagues, so UI correctly falls back to `rr_value.toFixed(1)` instead of always showing 1
- Leaderboard + league report now gate activity-specific points behind `points_only` formula; standard/simple = 1 point per approved/pending submission

## Affected files
- `api/leagues/[id]/submissions/route.ts`
- `api/leagues/[id]/my-submissions/route.ts`
- `api/leagues/[id]/my-team/submissions/route.ts`
- `lib/services/leaderboard-logic.ts`
- `lib/services/league-report.ts`

## Test plan
- [ ] Standard formula league: submission detail shows actual RR (e.g. 1.6 RR for 74 min gym)
- [ ] Points-only formula league: submission detail shows activity points
- [ ] Leaderboard scores correct for both formula types


adminmfl (approved):
Verified fix

PR #437, opened 2026-06-14
refactor(payments): Module 8 rate limiting, idempotency, webhook secu…

GitGuru-sudo:
**Summary**
Rate limiting, idempotency, webhook security fixes, proper Razorpay types, and payments page decomposition for Module 8.

**Changes Made**
- Added in-memory rate limiter (`src/lib/rate-limit.ts`) on all payment routes (10/5/5/5/30 req/min)
- Added idempotency support via `Idempotency-Key` header on `order` and `challenge-order` routes
- Created `POST /api/payments/webhook` route (was missing)
- Replaced all HMAC comparisons with `crypto.timingSafeEqual`
- Removed `err.message` leaks from all payment routes
- Stopped storing `razorpay_signature` in `updatePaymentStatus`
- Added proper `Window.Razorpay` type declarations in `src/types/razorpay.d.ts`
- Added Zod validation schemas for create-order, verify-payment, and webhook payloads
- Decomposed `payments/page.tsx` from 629→~140 lines into 4 components under `src/components/payments/`
- Created `GET /api/revenue` endpoint
- Fixed error message leak in `leagues/create/page.tsx:543`
- Added 22 tests (rate-limit: 6, payments validation: 12, webhook signature: 4)
- `pnpm build` passes on all routes

**How to Test**
1. Run `pnpm test` — all 22 new tests pass
2. Run `pnpm build` — compiles successfully
3. Hit any payment route rapidly to verify rate limiting kicks in after N requests
4. POST to `/api/payments/order` with an `Idempotency-Key` header twice with same key — second call returns cached response, doesn't create a duplicate order
5. Send a forged webhook to `/api/payments/webhook` — rejected with 401


blackhathckr (changes requested):
## Code Review — PR #437 (Module 8: Payments, Rate Limiting, Idempotency, Webhook Security)

### Must Fix

**1. SECURITY: `/api/revenue` route open to any authenticated user**
`revenue/route.ts` — Only checks `getAuthUser(req)`. The existing `/api/admin/revenue` requires `platform_role === 'admin'`. This new route exposes totalRevenue, monthlyRevenue, completedPayments, pendingPayments to any logged-in user. Privilege escalation. Add admin role check or remove (since admin route already exists).

**2. Idempotency check has TOCTOU race — no DB-level constraint**
`payments/order/route.ts` and `payments/challenge-order/route.ts` — SELECT then INSERT with no unique index on `notes->>idempotency_key`. Two concurrent requests can both pass the check and create duplicate payments. Add a partial unique index:
\`\`\`sql
CREATE UNIQUE INDEX payments_idempotency_key_uniq ON payments ((notes->>'idempotency_key')) WHERE notes->>'idempotency_key' IS NOT NULL;
\`\`\`

**3. `updatePaymentStatus` drops `razorpay_signature` parameter**
Signature changed from 4 args to 3 args, dropping `razorpay_signature`. Verify the DB column isn't NOT NULL and no downstream system depends on it.

**4. `payload: any` in webhook handler**
`payments-webhook.ts` — `webhookSchema` is defined in `lib/validations/payments.ts` but never actually used in the handler. Either apply the schema or type the payload properly.

### Should Fix

**5. In-memory rate limiter won't work across serverless instances**
`rate-limit.ts` stores counts in a module-level `Map`. On Vercel, each invocation may run in a different instance. Rate limiting will be ineffective in production.

**6. `setInterval` cleanup in `rate-limit.ts`**
The `setInterval(() => {...}, 60000)` at module scope has no `clearInterval`. In serverless this may never fire or cause issues.

**7. Idempotency-Key header required but client code not updated**
Order routes now require `Idempotency-Key` header, but league creation page doesn't send it. Every order creation will 400.

**8. Validation schemas defined but never used**
`lib/validations/payments.ts` defines 4 Zod schemas. None are imported by actual routes. Preview-price route actually replaced its Zod with manual checks (regression).

**9. `catch {}` blocks lose all diagnostic info**
Multiple routes changed from `catch (err) { console.error(...) }` to bare `catch { }`. Debugging production issues becomes impossible.

**10. Duplicate `rate-limit.ts` with PR #438**
Both PRs create the same file. Merge conflict incoming.

blackhathckr (changes requested):
## Re-review of PR #437 — Payments Rate Limiting, Idempotency, Webhook Security

Good progress addressing the original 10 review items. Most of the critical security issues have been resolved. Here is the item-by-item assessment, followed by one remaining must-fix.

---

### Original Items — Status

**1. Revenue endpoint auth (CRITICAL)** — FIXED
`/api/revenue/route.ts` now checks `authUser.platform_role !== 'admin'` and returns 403. This was the highest-priority issue and it has been properly addressed.

**2. Idempotency TOCTOU race (CRITICAL)** — FIXED
Migration `20260618_payments_idempotency_unique.sql` adds a partial unique index on `notes->>'idempotency_key'`. This closes the race at the DB level. Well done.

**3. `updatePaymentStatus` signature change** — FIXED
Signature is now `(orderId, paymentId, status, signature?)` with `razorpaySignature` as optional 4th param. The verify route and webhook handler both call it correctly with the new signature.

**4. `payload: any` in webhook handler** — FIXED
The webhook handler now uses `webhookSchema` from `lib/validations/payments.ts` to validate the payment event payload, and uses `eventEnvelopeSchema` for the outer envelope. Much better.

**5. In-memory rate limiter across serverless instances** — FIXED
`rate-limit.ts` now checks for Upstash/Vercel KV Redis env vars and uses Redis when available, falling back to in-memory. This is a good solution.

**6. `setInterval` cleanup** — FIXED
The old `setInterval` in `rateLimiter.ts` has been removed. The new `rate-limit.ts` uses per-request window expiry instead. `rateLimiter.ts` is now a thin re-export shim.

**7. Idempotency-Key header not sent by client** — PARTIALLY FIXED
The league creation page (`leagues/create/page.tsx`) now generates and sends the `Idempotency-Key` header. However, the `challenge-order` route also requires this header (line 853 of the diff returns 400 without it), and no client-side code for challenge ordering was updated. If there is a client caller for `challenge-order`, it will break. If there is no client caller yet, this is fine — but please confirm.

**8. Validation schemas defined but unused** — FIXED
`verifyPaymentSchema` is now used in the verify route. `webhookSchema` is used in the webhook handler. `previewPriceSchema` is used in preview-price. `createOrderSchema` is defined with tests but not actually imported by any route — acceptable since the order routes use manual validation that matches the schema's intent.

**9. `catch {}` blocks losing diagnostic info** — MOSTLY FIXED
Server-side catch blocks throughout the codebase now capture and log the error (e.g., `catch (err) { console.error(...) }`). The two remaining bare `catch {}` in client-side React (`leagues/create/page.tsx` line 544, `payments/page.tsx` line 325) are a minor regression (lost `console.error`) but acceptable for client code since errors are surfaced via `setError()`. The bare `catch {}` blocks in `timingSafeEqual` wrappers are fine — they intentionally swallow length-mismatch errors.

**10. Duplicate `rate-limit.ts` with PR #438** — ADDRESSED
`rateLimiter.ts` is now a re-export shim pointing to the canonical `rate-limit.ts`. This should resolve the merge conflict scenario cleanly.

---

### NEW ISSUE — Must Fix

**`legacyRedisIncr` return type mismatch in `checkRateLimit`**

In `src/lib/rate-limit.ts`, the function `legacyRedisIncr` returns `{ allowed, remaining, reset }`, but `checkRateLimit` destructures its result as `{ count, ttl }` (line ~2336 in the diff). Since `count` and `ttl` don't exist on the returned object, they will be `undefined`. This means:
- `count <= limit` → `undefined <= N` → `false` → `allowed = false` → rate limiter always blocks when Redis is available
- The `catch` block will swallow this silently and fall back to in-memory, so it won't crash, but the Redis path is completely dead code

Additionally, `legacyRedisIncr` itself has bad logic: `allowed: count <= 0` (always false since INCR starts at 1) and `remaining: Math.max(0, 0 - count)` (always 0). This function should either:
- Return `{ count, ttl }` directly (matching how it's consumed), or
- Be removed entirely since the caller re-derives `allowed` and `remaining` from `count` and `limit` anyway

Fix: change `legacyRedisIncr` to return `{ count, ttl }` like `redisIncr` does, matching the caller's destructuring.

---

### Summary

9 of the original 10 items are fixed. Item 7 (challenge-order client) needs confirmation. One new bug was introduced in the legacy rate limiter Redis path. Requesting changes for the `legacyRedisIncr` return mismatch — it is a real bug that silently disables Redis-backed rate limiting for all legacy auth routes.

— reviewed by @blackhathckr

blackhathckr (approved):
All previous feedback addressed. Zero any casts, legacyRedisIncr fixed. LGTM.

PR #492, opened 2026-06-23
fix(challenges): show real total_points on configure-challenges cards…

GitGuru-sudo:
Here's the filled PR description:

---

## Summary
Fixes BUG-012 — Configure Challenges cards always show "0 pts" because the demo seed never set `leagueschallenges.total_points`. Seeds a realistic points ceiling per demo challenge, hardens the card to distinguish "no ceiling" from "0 points", and removes a dead API fallback.

## Related Issue
Closes BUG-012

## Type of Change
- [x] Bug fix

## Changes Made
- `supabase/seeds/demo-leagues.sql` — Added `UPDATE` after each `INSERT INTO public.leagueschallenges` (20 total across 3 demo leagues) to seed `total_points` with values from 100–300, scaling from early challenges to finales
- `src/app/(app)/leagues/[id]/configure-challenges/page.tsx` — Replaced bare `{challenge.total_points} pts` with conditional: shows "Points not set" when the ceiling is 0/null, otherwise shows the real points value
- `src/app/api/leagues/[id]/challenges/route.ts` — Dropped the `template?.total_points` fallback (column doesn't exist on `specialchallenges`, so it was always `undefined` — dead code that misleads readers)

## How to Test
1. Open `/leagues/[id]/configure-challenges` for any demo league — each card shows a real points value (e.g. "100 pts"), not "0 pts"
2. Edit a challenge, clear Total Points, save — card reads "Points not set"
3. Edit again, set 250, save — card reads "250 pts"
4. "Publish Scores" still works (it reads `awarded_points`, unaffected)
5. `pnpm db:seed` on a fresh DB — demo challenges have non-zero ceilings (idempotent)

## Checklist


blackhathckr (approved):
Good fix. UI now shows 'Points not set' instead of '0 pts', dead template fallback removed, challenge points seeded in demo data. LGTM.

PR #505, opened 2026-06-25
feat(reports): host league activity report — daily/weekly/custom with…

GitGuru-sudo:
## Summary

Host/governor-only league-wide activity report with three range modes (daily/weekly/custom) and two export formats (CSV/PDF).

## Related Issue

Closes #494

## Changes Made

- New `host-report-collector.ts` — league-wide, all-status effortentry fetch with date window resolution
- New `host-report-aggregator.ts` — pure functions: daily rows (sorted team→player), weekly per-player + per-team summaries (Mon–Sun), league summary (HR-positive, streaks, participation %)
- New `host-report-csv.ts` — RFC-4180 CSV serializer with per-section column maps
- New `host-report-pdf.tsx` — `@react-pdf/renderer` component with three layouts
- New `host-report-dialog.tsx` — UI with presets (Today/This Week/Last 7 Days/This Month/Custom) + CSV/PDF selector
- Edited `route.ts` — added role-gated `range` branch before legacy per-user path
- Edited `header-actions.tsx` + `page.tsx` — render "Download Report" button only for host/governor
- 29 unit tests covering aggregation math, sort order, CSV escaping, empty states

## How to Test

1. As a **host**, open a league dashboard → click "Download Report" → try each preset with CSV and PDF
2. Verify CSV opens in Excel with correct columns, sorted team→player, all statuses shown
3. Verify player accounts get 403 when hitting the API directly
4. Verify active (non-completed) leagues work for all range modes


blackhathckr (changes requested):
Multiple as any casts need fixing:
- cell(value: any, ..., extraStyle: any) — type these params
- standings: any — add proper interface
- .map((m: any) => ...) — type the Supabase row results
- (m as any).users?.username — type the join
- const r = row as any — type the row

This is a large new feature (2000 lines). Clean up the types before merge. Also note: do NOT merge until after Friday production release.

blackhathckr (changes requested):
## Review of PR #505 — Host League Activity Report

### Previous Review Comments — Status

The June 26 review requested fixing multiple `as any` casts:

| Previous Comment | Status |
|---|---|
| `cell(value: any, ..., extraStyle: any)` — type these params | **Fixed** — `cell` now uses `string \| number \| null \| undefined` for value and `CellStyle \| CellStyle[]` for extraStyle (`host-report-pdf.tsx` ~line 96) |
| `standings: any` — add proper interface | **Fixed** — typed as `LeaderboardData \| null` with a proper interface in `host-report-collector.ts` (~line 107) |
| `.map((m: any) => ...)` — type the Supabase row results | **Fixed** — explicit row interfaces `MemberQueryRow`, `TeamQueryRow`, `LeagueActivityQueryRow`, `EffortEntryRow` added; `.returns<T>()` used on all queries |
| `(m as any).users?.username` — type the join | **Fixed** — `MemberQueryRow` includes `users: { username: string } \| null` |
| `const r = row as any` — type the row | **Fixed** — row interfaces remove the need for this cast |

**All five previous comments have been addressed.** The commit `275e903` (Jun 28) specifically targets these.

---

### New Issues Found

#### 1. Remaining `as any` in route.ts (~line 72 of diff / approx line 83 in file)
```ts
React.createElement(HostReportPDF, { kind: range, data: collected }) as any
```
**Verdict: Acceptable** — this matches the existing pattern for `renderToBuffer` used with `CertificatePDF` and `LeagueReportPDF` in the same file. This is a known `@react-pdf/renderer` typing limitation. Not blocking.

#### 2. [Bug] No validation on `range` query param — `route.ts` ~line 42-56
`range` comes from `searchParams.get('range')` (type `string`) and is cast to `'daily' | 'weekly' | 'custom'` without runtime validation. If a caller passes `range=foo`, `resolveWindow` will silently treat it as `custom` (falls through to the else branch). More importantly, `kind: range` on line 70 passes the raw string to `HostReportPDF` which expects `'daily' | 'weekly' | 'custom'` — this is a type error that TypeScript may not catch due to the `as any` on `React.createElement`.

**Recommended fix**: Add a guard:
```ts
const validRanges = ['daily', 'weekly', 'custom'] as const;
if (!validRanges.includes(range as any)) {
  return NextResponse.json({ error: 'Invalid range' }, { status: 400 });
}
const validatedRange = range as (typeof validRanges)[number];
```

#### 3. [Bug] Empty `memberIds` array passed to `.in()` — `host-report-collector.ts` ~line 245
If a league has zero members, `memberIds` will be `[]`. Passing an empty array to Supabase `.in('league_member_id', [])` can return unexpected results depending on PostgREST version (some return all rows, some return none). Should guard with:
```ts
if (memberIds.length === 0) {
  return { league: ..., members: [], teams: [], submissions: [], standings: null, window };
}
```

#### 4. [Nit] Unused `standings` fetch — `host-report-collector.ts` ~line 276
`getLeaderboardData(leagueId)` is called on every report request, but `standings` is never consumed by any aggregation function (`buildDailyRows`, `buildWeeklySummary`, `buildLeagueSummary`). This is a wasted API call that adds latency to every report generation. Either remove it or document why it's being collected for future use.

#### 5. [Nit] Duplicated `calculateStreak` — `host-report-aggregator.ts` ~lines 173 and 345
The `calculateStreak` function is copy-pasted identically in both `buildWeeklySummary` and `buildLeagueSummary`. Extract it to a shared module-level function.

#### 6. [Nit] O(n²) `.find()` inside loops — `host-report-aggregator.ts` ~lines 119, 130, 301, 404
Six calls to `input.members.find()` inside loops over submissions. For large leagues with many submissions, this is O(members × submissions). Build a `Map<string, HostMemberRow>` keyed by `league_member_id` at the top of each function (like the collector already does in `memberMap` at line 250).

#### 7. [Nit] React lint: `useMemo` dependency array missing refs — `host-report-dialog.tsx` ~line 246
`todayDate`, `firstOfMonth`, `getMonday`, `getSunday` are used inside the `queryParams` `useMemo` but not listed in its dependency array `[preset, today, customStart, customEnd]`. While these are stable values (date doesn't change mid-session), this will trigger `react-hooks/exhaustive-deps` warnings. Either memoize those values or add them to deps.

#### 8. [Nit] Format radio buttons not wrapped in `RadioGroup` — `host-report-dialog.tsx` ~lines 215-236
The CSV/PDF format selector uses standalone `RadioGroupItem` components with manual `checked`/`onClick` props instead of being wrapped in a `RadioGroup` with `value`/`onValueChange`. This works but is not idiomatic Radix UI usage and may break accessibility (arrow key navigation between options won't work).

#### 9. [Nit] Test type mismatch — `host-report-aggregator.test.ts`
Test mock data passes `distance: '5'` (string) in many places, but `HostSubmissionRow.distance` is typed as `number | null`. The code in `buildDailyRows` does `Number(sub.distance)` which suggests the DB column is actually a text/string type. If so, the `HostSubmissionRow` and `EffortEntryRow` interfaces should type `distance` as `string | null` instead. If the DB column is numeric, then the `Number()` conversion is unnecessary. Whichever it is, the type and the code should agree.

---

### Summary

| Category | Count |
|---|---|
| Previous issues fixed | 5/5 |
| New bugs (should fix) | 2 (#2, #3) |
| New nits (nice to fix) | 5 (#4-#8) |
| Test type issue | 1 (#9) |

**Verdict: Request changes** — issues #2 (range validation) and #3 (empty memberIds) are correctness bugs that should be fixed before merge. The nits are non-blocking but worth addressing in this PR since it's new code.

blackhathckr (changes requested):
Good work fixing the `as any` casts from the last review — all 5 are properly typed now with `MemberQueryRow`, `TeamQueryRow`, `.returns<T>()`, etc.

Two issues to fix before approval:

### Bugs

1. **No validation on `range` query param** (`src/app/api/leagues/[id]/report/route.ts`): The raw string is cast to `'daily' | 'weekly' | 'custom'` without validation. Invalid values silently fall through to `custom` behavior. Add a guard that returns 400 for invalid range values.

2. **Empty `memberIds` passed to `.in()`** (`src/lib/services/reports/host-report-collector.ts` ~line 245): If a league has zero members, `supabase.in('league_member_id', [])` has undefined behavior across PostgREST versions. Add an early return guard when `memberIds` is empty.

### Nits (fix while you're at it)

- `standings` is fetched but never used in the report output — remove the fetch to save latency
- `calculateStreak` is duplicated — import from the existing util instead
- O(n²) `.find()` inside loops — use a `Map` for lookups
- `useMemo` dependency array in `host-report-dialog.tsx` is incomplete
- Format radio buttons not in a `RadioGroup` — accessibility issue

blackhathckr (changes requested):
## Review — PR #505: Activity Report (Fix Commit)

Good progress — all previously requested fixes are addressed:
- ✅ `VALID_RANGES` validation guard added
- ✅ Early return on empty `memberIds`  
- ✅ `standings` fetch removed (hardcoded to `null`)
- ✅ O(n) Map lookups replacing O(n²) `.find()` calls
- ✅ `calculateStreak` hoisted to module-level
- ✅ Radio group accessibility in dialog
- ✅ Helper functions moved outside component

### Remaining Issues

1. **`format` parameter not validated** (`src/app/api/leagues/[id]/report/route.ts`)
   `range` is validated against `VALID_RANGES`, but `format` is not. It silently defaults to CSV if an unknown value like `format=xml` is passed. Add a `VALID_FORMATS` guard similar to `VALID_RANGES`:
   ```ts
   const VALID_FORMATS = new Set(['pdf', 'csv']);
   if (!VALID_FORMATS.has(format)) return NextResponse.json({ error: 'Invalid format' }, { status: 400 });
   ```

2. **`format` variable shadowed with different defaults**
   The outer scope defaults to `'pdf'`, the inner host-report scope defaults to `'csv'`. This is confusing. Rename the inner one to `hostFormat` or restructure so the host path early-returns before the outer declaration.

3. **"Activities" metric is misleading** (`host-report-aggregator.ts`)
   In `buildWeeklySummary`, `activities` = `entry.activities.size` counts **distinct activity types** (Running, Cycling), not total workout sessions. Hosts will likely interpret "Activities" as total workouts. Either rename to `activityTypes` / `uniqueActivities`, or change the calculation to count total entries.

4. **`activityNameMap` lookup broken for custom activities** (`host-report-collector.ts`)
   Custom activities use `custom_activity_id` as the map key, but the lookup uses `workout_type` (a name string). The custom activity branch of the map is effectively dead code. Either key by `workout_type` consistently, or look up by both ID and name.

### Nit
5. `useMemo(() => formatYMD(new Date()), [])` — stale if dialog stays open past midnight. Minor, but add a comment or skip memoization since the computation is trivial.

Fix items 1-4, then this is good to go.

blackhathckr (approved):
## Review — PR #505: Activity Report (Latest Fix)

All previously requested changes are now addressed:

- ✅ `VALID_FORMATS` guard with 400 response for invalid format
- ✅ `format` → `hostFormat` — no more variable shadowing
- ✅ `activities` → `activityTypes` in type, aggregator, CSV headers, and PDF headers
- ✅ Custom activity map now keyed by both name and ID with clear comment
- ✅ `useMemo` staleness documented with comment
- ✅ Code formatting cleaned up

LGTM. Good work addressing all the feedback across multiple review rounds.

PR #510, opened 2026-06-26
feat: add submission limit configuration to leagues and implement overwrite logic for daily entry limits

dhjr:
## Summary
This PR fixes a bug where league-level daily submission limits were not being properly saved to the database when updated by the host, and resolves an issue where overwriting an existing activity falsely triggered the daily maximum limit error.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] UI/UX improvement
- [ ] Documentation

## Changes Made
- **Settings Persistence:** Added `min_submissions_per_day` and `max_submissions_per_day` to the `allowedUpdates` whitelist in `league-mutation-service.ts` so that configuration changes correctly persist for active and launched leagues.
- **Overwrite Enforcement Fix:** Updated the submission limit logic in `entries/upsert/route.ts` to correctly handle same-activity resubmissions. The system now allows users to overwrite their existing entries without falsely counting the updated entry as an additional submission against the daily limit.

## How to Test
1. As a league host, navigate to League Settings -> Activities.
2. Update the "Max Submissions Per Day" to `2` and click **Save**.
3. Refresh the page to verify the changes have successfully persisted to the database.
4. As a player, log an activity to reach your newly set daily limit.
5. Attempt to log the **exact same** activity type again. You should receive the overwrite prompt and be able to successfully update your entry without hitting the limit blocker.
6. Attempt to log a **different** activity type that pushes you over the daily limit. You should be correctly hard-blocked with the `(2/day)` limit error.

## Checklist


blackhathckr (commented):
## Review

Good fix — two real bugs addressed here:

1. **Settings persistence**: `min_submissions_per_day` and `max_submissions_per_day` were missing from the `allowedUpdates` whitelist in `league-mutation-service.ts`, so hosts could change the values in the UI but they'd never persist. Good catch.

2. **Overwrite logic**: When a user resubmits the same activity type and the daily limit is already reached, the old code would block the overwrite. The new logic correctly checks if there's an existing entry of the same `workout_type` and allows the overwrite path to proceed. Different activity type still gets hard-blocked. Correct.

**One nit:**

`src/app/api/entries/upsert/route.ts` — line with `(r: any) => r.workout_type === workout_type`:
The query already does `.select('id, status, workout_type, type')` so the shape is known. Instead of `(r: any)`, type it properly:
```ts
const existingSameType = activeEntries.find(
  (r: { id: string; status: string; workout_type: string; type: string }) =>
    r.workout_type === workout_type,
);
```
Or extract an inline type/interface. The existing `(row: any)` on the filter above has the same problem — consider fixing both while you're here.

Also, the `tour-steps-table.tsx` change (adding missing newline + closing brace) looks like it was an unrelated fix that got bundled in. Not a problem, just noting it.

blackhathckr (commented):
## Review

Good fix — two real bugs addressed here:

1. **Settings persistence**: min/max submissions_per_day were missing from the allowedUpdates whitelist in league-mutation-service.ts, so hosts could change the values in the UI but they never persisted. Good catch.

2. **Overwrite logic**: When a user resubmits the same activity type and the daily limit is already reached, the old code would block the overwrite. The new logic correctly checks if there is an existing entry of the same workout_type and allows the overwrite path to proceed. Different activity type still gets hard-blocked. Correct.

**One nit:**

src/app/api/entries/upsert/route.ts — the line with (r: any) => r.workout_type === workout_type:
The query already does .select('id, status, workout_type, type') so the shape is known. Replace (r: any) with a proper type. The existing (row: any) on the filter above has the same issue — fix both while you are here.

Also, the tour-steps-table.tsx change (adding missing newline + closing brace) looks like an unrelated fix bundled in. Not a problem, just noting it.

blackhathckr (approved):
Type nit fixed — removed unnecessary `any` casts. LGTM.

PR #511, opened 2026-06-26
feat: auto-redirect users to their league if they belong to only one total

dhjr:
## Summary
Updates the dashboard and leagues routing logic to automatically redirect users directly to their league page if they only belong to a single league, regardless of the league's current phase or status (draft, active, scheduled, etc.).

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] UI/UX improvement
- [ ] Documentation

## Changes Made
- **`src/app/(app)/dashboard/page.tsx`**: Updated the server-side logic to immediately redirect users with `leagues.length === 1` to their single league, bypassing the active status check.
- **`src/app/(app)/leagues/page.tsx`**: Mirrored this behavior in the client-side `useEffect` hook so that clicking the "MFL Dashboard" navigation button handles the single-league edge case consistently.

## How to Test
1. Log in to an account that belongs to exactly one league (ensure the league is in a non-active phase like `draft` or `launched`).
2. Click the "MFL Dashboard" link in the navigation sidebar (or the MFL logo).
3. Verify that you are immediately redirected to `/leagues/[league_id]` instead of hanging on the `/dashboard` or `/leagues` overview page.

## Checklist


blackhathckr (commented):
## Review

Clean and simple. Two files changed, both doing the same thing — if user belongs to exactly 1 league (any status), skip the league selection page and go straight to it.

**dashboard/page.tsx (SSR):** server-side redirect before rendering — good.
**leagues/page.tsx (client):** client-side useEffect redirect — mirrors the server logic correctly.

One concern:

The old behavior was: redirect only if the user has exactly 1 **active** league. The new behavior redirects even if the single league is in draft/scheduled/ended state. Is this intentional? If the user's only league is ended, they land on an ended league page with no way to go back to the leagues list to create/join a new one (since /leagues auto-redirects them back). Think about whether there should be at least a back/exit affordance on the league page.

Otherwise, the removed DumbbellLoading and PwaHeroInstall imports — these were unused? Verify they are not referenced elsewhere in the file before removing.

Looks good overall. Minor edge case to consider above.

blackhathckr (approved):
Fixed the ended-league redirect loop edge case. LGTM.

PR #517, opened 2026-06-28
fix: populate Team dropdown from teamleagues instead of inferred from…

GitGuru-sudo:
## Summary
Team dropdown in Manual Entry now sources its options from the `teamleagues` table (all teams in the league) instead of inferring them from the filtered member list. Previously, if no players were assigned to a team, `teamOptions` was empty and the dropdown showed nothing but "All teams".

## Related Issue
Closes # —

## Type of Change
- [x] Bug fix

## Changes Made
- **`route.ts`**: Added a query to `teamleagues` → `teams` to fetch the league's actual teams and return them as `teams` in the GET response alongside `data` (members).
- **`page.tsx`**: `teamOptions` now derives from the response's `teams` array instead of extracting team names from members. Player filtering by team still uses the member's `team_name` — both sources originate from the same `teams` table, so they match exactly.

## How to Test
1. Create a league with teams but leave players unassigned to any team.
2. Navigate to Manual Entry (`/leagues/[id]/manual-entry`).
3. **Team dropdown should show all league team names** (previously empty).
4. Select a team — player dropdown shows correctly filtered members (if any are assigned) or is empty (correct state).
5. Assign some players to teams, repeat — team dropdown still shows all teams, player filtering works correctly.
6. Verified manually on local dev server — logic works correctly for both scenarios (teams with and without assigned members).

## Checklist


blackhathckr (changes requested):
## Review: Fix Team dropdown in manual entry

Core approach is correct — querying \`teamleagues\` is the right source of truth for league-team associations. Shows all teams even if they have no assigned members.

### Issues

**1. Silent error swallowing (must fix)**
\`src/app/api/leagues/[id]/manual-entry/route.ts\` — When \`teamleagues\` query fails, error is logged but response returns \`{ success: true, teams: [] }\`. User sees no teams and no error indication. Before this PR, teams were at least inferred from members.

**Fix:** Either return a 500 when the teams query fails, or fall back to inferring teams from members when \`teams\` is empty.

**2. \`normalizedTeamName\` inconsistency (must fix)**
\`src/app/(app)/leagues/[id]/manual-entry/page.tsx\` — Removed from \`teamOptions\` (now uses raw \`t.team_name\`) but still applied in \`filteredMembers\`. If team name has trailing whitespace, the dropdown value and filter comparison could diverge.

**Fix:** Remove \`normalizedTeamName\` from \`filteredMembers\` too — the server now guarantees non-null, clean team names via the \`.filter()\`.

**3. NIT: Type guard missing**
The \`.filter((t) => t.team_name !== null)\` doesn't narrow the type because of the \`any\` mapping. Use: \`.filter((t): t is { team_id: string; team_name: string } => t.team_name !== null)\`

**4. NIT: PR body says \`Closes # —\` with no actual issue number.** Remove or link the correct issue.

### What looks good
- Auth checks intact (getAuthUser + role check)
- Additive response shape change (won't break existing consumers)
- Correct scoping with \`.eq('league_id', leagueId)\`

Fix all issues, then re-request review.

blackhathckr (approved):
## Re-review: PR #517 — Fix Team dropdown in manual entry

All 4 issues from the original review are fixed:

1. **Silent error swallowing** — ✅ Returns `500` with `{ error: 'Failed to fetch teams' }` when `teamleagues` query fails
2. **`normalizedTeamName` inconsistency** — ✅ `normalizedTeamName` function completely removed. Both `teamOptions` and `filteredMembers` now use raw `team_name` from server
3. **Type guard missing** — ✅ Proper type guard with `ManualEntryTeam` interface: `.filter((t): t is ManualEntryTeam => t.team_name !== null)`
4. **Added proper interfaces** — `TeamLeagueRow` and `ManualEntryTeam` for type safety

Client-side correctly consumes `json.teams` from the updated response shape with new `TeamOption` interface.

PR #520, opened 2026-06-29
Feat/timezone cron

### LINKED ISSUE #465: feat: Store user timezone and use per-user timezone in auto-rest-day cron
## Background

The auto-rest-day cron (`api/cron/auto-rest-day/route.ts`) currently uses hardcoded IST (UTC+5:30) for all users. When members travel abroad (e.g., to US timezones), they receive an auto rest day before their day is actually over, because midnight IST is early afternoon in the US.

**Decision**: Use the timezone from the user's last submission/login to determine "yesterday" per-user.

### 1. Store user timezone on every submission

In `api/entries/upsert/route.ts`, we already receive `ianaTimezone` (line 90). After a successful submission, persist this to the user's profile:

```
users.last_known_timezone = ianaTimezone  (e.g., 'America/New_York')
```

- Add a `last_known_timezone` column to the `users` table (text, nullable, default null)
- Update it on every successful submission (non-blocking, fire-and-forget)
- Fallback: if null, use IST (existing behavior)

### 2. Modify auto-rest-day cron to use per-user timezone

Currently the cron computes a single "yesterday" for all users using IST. Change to:

- For each member being processed, look up `users.last_known_timezone`
- Compute "yesterday" using that timezone
- If timezone is null, fall back to IST
- Only assign auto rest day if yesterday has actually ended in the user's timezone

### 3. Auto-rest overwrite already works (no change needed)

The upsert route already replaces auto-assigned rest days with workout submissions (line 1195). This is the safety net — if a user gets an auto-rest but then submits a workout, it auto-corrects.

## Files to modify

- `src/app/api/entries/upsert/route.ts` — save `ianaTimezone` to user profile after successful submission
- `src/app/api/cron/auto-rest-day/route.ts` — use per-user timezone for "yesterday" calculation
- DB migration: add `last_known_timezone` column to `users` table

## Acceptance criteria

- [ ] `last_known_timezone` column exists on `users` table
- [ ] Every successful submission updates the user's `last_known_timezone`
- [ ] Auto-rest-day cron uses per-user timezone (falls back to IST if null)
- [ ] Users in non-IST timezones don't get premature auto rest days
- [ ] Existing behavior unchanged for users without stored timezone

dhjr:
## Summary
This PR shifts the auto-rest-day cron job from a global IST-based calculation to a per-user timezone model. By persisting the user's timezone on every successful submission, the system now accurately assigns auto-rest days only when a user's local day has genuinely ended.

## Related Issue
Closes #465

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] UI/UX improvement
- [ ] Documentation

## Changes Made
- Added a `last_known_timezone` column to the `users` table via migration.
- Updated the `api/entries/upsert` route to capture `ianaTimezone` and asynchronously persist it to the user's profile on every successful submission.
- Refactored the `auto-rest-day` cron job to fetch each member's `last_known_timezone` and compute a localized "yesterday" using `Intl.DateTimeFormat`, effectively preventing premature rest day assignments.

## How to Test
1. Submit an activity using an account with a non-IST timezone (e.g., `America/New_York`).
2. Verify that `last_known_timezone` is updated in the `users` table for that account.
3. Manually trigger the `auto-rest-day` cron job.
4. Verify that the cron successfully calculates the correct localized "yesterday" and only assigns an auto-rest day if that specific date has elapsed in the user's timezone.
5. Confirm that users with no timezone stored gracefully fall back to IST behavior.

## Checklist


blackhathckr (changes requested):
Review posted - see details below

blackhathckr (changes requested):
## Review: Timezone Cron

Good approach — per-user IANA timezone via \`Intl.DateTimeFormat\` instead of hardcoded IST. DST handled correctly by IANA identifiers. Fallback chain (user tz → IST → UTC) is reasonable.

### Must fix

**1. BUG: \`(member as any)\` unsafe cast**
\`src/app/api/cron/auto-rest-day/route.ts\` — suppresses type checking on the joined query result. Define a proper type for the member with \`last_known_timezone\`.

**2. BUG: Fire-and-forget promise missing \`.catch()\`**
\`src/app/api/entries/upsert/route.ts\` (two locations) — the Supabase \`.update()\` call has \`.then()\` but no \`.catch()\`. Unhandled rejection if network/runtime error. Also duplicated — extract to a helper.

**3. NIT: No validation of \`ianaTimezone\` before DB write**
Same file — only checks \`typeof === 'string' && .trim()\`. Client could send \`"foo bar"\` and pollute the DB. Validate with \`Intl.DateTimeFormat(undefined, { timeZone: tz })\` in a try/catch.

**4. NIT: Stale comments**
- File header (lines 1-13) still says "Calculate their local yesterday based on IST (UTC+5:30)" and "By running at midnight IST"
- Section comment (~line 34) says "timezone offset" but function now takes IANA string

**5. NIT: Unrelated \`.agents\` in \`.gitignore\`** — should be a separate commit.

Fix all, then re-request review.

blackhathckr (changes requested):
## Overview

Shifts the auto-rest-day cron from a global IST offset to a per-user IANA timezone model. Persists `last_known_timezone` on every submission and uses `Intl.DateTimeFormat` to compute each user's local "yesterday" — DST-safe and correct across all zones.

**Real logic changes are in 3 files** (~120 LoC):

- `supabase/migrations/20260629000001_add_last_known_timezone.sql` — adds `last_known_timezone text` (idempotent `IF NOT EXISTS`, nullable)
- `src/app/api/cron/auto-rest-day/route.ts` — replaces offset math with `Intl.DateTimeFormat` + falls back to `Asia/Kolkata`
- `src/app/api/entries/upsert/route.ts` — adds `updateUserTimezone()` helper called on 2 success paths

The other 7 files (`ai-nudges/route.ts`, `auto-activate-leagues/route.ts`, `league-lifecycle.ts`, `nudge-*.ts`, `nudge-engine.test.ts`, `profile-header.tsx`) are **pure Prettier reformats + one unrelated unused-import removal** — no functional change.

## Blockers

### 1. Supabase FK relationship shape may be an array, not an object

`src/app/api/cron/auto-rest-day/route.ts` around L239 and L269:

```ts
.select(`league_member_id, user_id, users ( last_known_timezone )`)
...
const typedMember = member as {
  ...
  users: { last_known_timezone: string | null } | null;
};
```

This casts `users` as a single object, but Supabase's default relational resolution can return an **array** for the join, in which case `typedMember.users?.last_known_timezone` silently reads `undefined` and every user quietly falls back to IST — defeating the whole PR.

We already fixed this exact pattern in #531 (`fix/ssr-role-resolution`) on 2026-07-01. Apply the same pattern here:

- Force object shape via inner join: `users!inner ( last_known_timezone )`, or
- Handle both shapes explicitly: `const u = Array.isArray(m.users) ? m.users[0] : m.users`

Please verify against real Supabase output before assuming this cast is safe.

### 2. Fire-and-forget DB write can be killed by serverless termination

`src/app/api/entries/upsert/route.ts` — `updateUserTimezone` starts the DB update without `await`, then the handler returns `NextResponse.json(...)` immediately. On Vercel serverless, the runtime does **not** guarantee pending promises complete after the response is sent unless registered via `waitUntil`.

Fix options (any of):

- `await` the update (adds one roundtrip to every submission; probably fine, submissions aren't hot-path)
- Use Next 15 `after()`:
  ```ts
  import { after } from 'next/server';
  after(() => supabase.from('users').update(...));
  ```
- Vercel `waitUntil(promise)` from `@vercel/functions`

Silent data loss here would look identical to blocker #1 — everyone silently falls back to IST.

## Nits

- **Diff pollution** — 7 files are Prettier-only or unrelated (unused `User` icon removed from `profile-header.tsx`). Please revert those and either (a) land them separately or (b) run prettier on the whole repo in its own PR. Right now roughly 25% of the diff isn't part of this feature.
- **No tests added** — `getYesterdayInTimezone` is on the critical rest-day path and trivial to unit test:
  ```ts
  getYesterdayInTimezone('America/New_York')   // valid IANA
  getYesterdayInTimezone(null)                 // fallback to IST
  getYesterdayInTimezone('Not/A/Zone')         // invalid, catches, recurses to IST
  ```
- **Absolute fallback semantics** — inside the `catch`, when `tz === 'Asia/Kolkata'` you fall through to a UTC-based yesterday, not IST-based. Won't happen in practice but the comment "Absolute fallback" is misleading; either drop that branch or make it IST-consistent.
- **No debounce on write** — every submission writes the timezone even if unchanged. Not worth reading-then-writing to avoid it, but worth a comment acknowledging the design choice.

## Positive

- IANA over offsets is the right call — DST handled natively by `Intl`.
- Write-side validation via `Intl.DateTimeFormat(undefined, { timeZone: x })` — clean, throws on invalid strings.
- `unknown` typing on `ianaTimezone` param + typeof check — good defensive coding.
- Idempotent migration (`IF NOT EXISTS`) — safe to re-run.
- Cron cadence math checks out: cron fires at 18:30 UTC daily; for every IANA zone, "yesterday-local" has fully elapsed by that time.

## Recommendation

**Request changes.** Blockers #1 (FK shape) and #2 (serverless termination) are both silent-failure modes that look identical — everyone falls back to IST and no one notices. Fix those two + strip the Prettier noise, then this is ready to approve.

blackhathckr (dismissed):
## Re-review: Timezone CRON (post-fix commits)

Both blockers from the Jul 6 review are resolved:

**Blocker #1 (FK shape)** — Fixed. `users` is now typed as `object | array | null` with `Array.isArray` guard before access. No more silent fallback to IST.

**Blocker #2 (fire-and-forget)** — Fixed. Both `updateUserTimezone` call sites are now `await`ed. No more risk of serverless termination killing the write.

**Additional fixes addressed:**
- Stale IST comments removed from file header
- IANA validation via `Intl.DateTimeFormat` try/catch before DB write
- Unit tests added for `getYesterdayInTimezone` (valid, null, invalid)
- Debounce design choice documented in comment

**Remaining nit (non-blocking):** 7 files are still Prettier-only reformats. Ideally strip those into a separate commit, but not worth blocking over.

LGTM — approve.

blackhathckr (changes requested):
Dismissing previous approval — nits must be fixed before approval.

Core timezone changes are solid (see previous review). One remaining item:

1. **7 Prettier-only file reformats** — these files have no functional changes, just formatting. Strip them into a separate commit or remove from this PR to keep the diff clean and reviewable. The actual timezone logic is good — don't bury it under unrelated formatting noise.

blackhathckr (dismissed):
Prettier-only changes reverted. PR diff is now clean — only functional timezone changes remain. LGTM.

blackhathckr (changes requested):
Prettier revert is good. One remaining issue:

**`console.error` in `updateUserTimezone`** (`src/app/api/entries/upsert/route.ts`):

```ts
if (error) console.error('Failed to update timezone:', error);
```

This is new code added by this PR. Use `logger.error` from `@/lib/logger`.

blackhathckr (changes requested):
Logger fix for console.error is good. One more in the same function:

**`console.warn` in `updateUserTimezone`** (src/app/api/entries/upsert/route.ts):

```ts
} catch (e) {
  console.warn('Invalid IANA timezone provided:', ianaTimezone);
}
```

This is new code from this PR. Use `logger.warn` from `@/lib/logger`.

blackhathckr (approved):
Verified commit `babf1ce`: the last remaining item (`console.warn` in `updateUserTimezone`) is now `logger.warn` with structured metadata. All previous items across the review history are resolved — FK shape guard, awaited timezone updates, Prettier revert, logger for console.error. Approving.

PR #530, opened 2026-07-01
feat(admin): co-host assignment on any league (#526)

GitGuru-sudo:
Admin panel multi-host support: from the league management view, an admin can assign any registered user as co-host (search by email), remove co-hosts, and see the owner + co-host list. New admin API - GET/POST /api/admin/leagues/[leagueId]/co-host and DELETE .../co-host/[userId] - reuses the assignedrolesforleague 'host' role (no schema change). Guarded by platform_role=admin (any league, not just created_by); the original creator can never be removed. Core multi-host domain logic ships separately in #525.

## Summary

Admin co-host management panel — MFL support team can assign/remove co-hosts on any league via new API routes and a "Manage Hosts" dialog in the league table actions dropdown.

## Related Issue

Closes #526

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] UI/UX improvement
- [ ] Documentation

## Changes Made

- New `src/lib/services/admin/co-host.ts` — shared `getHostRoleId()` helper
- New `GET/POST /api/admin/leagues/[leagueId]/co-host` — list owner + co-hosts, assign a co-host (reuses `assignRoleToUser`)
- New `DELETE /api/admin/leagues/[leagueId]/co-host/[userId]` — remove co-host (reuses `removeRoleFromUser`), guarded against removing `created_by` (409)
- New `ManageHostsDialog` — dialog with owner badge (Crown), co-host badges (Shield), user search via existing `/api/admin/users?search=`, add/remove with loading states
- Edited `leagues-table.tsx` — "Manage Hosts" dropdown item between "Login as Host" and "Edit", dialog wired with open/close state

## How to Test

1. Login as admin → `/admin/leagues`
2. Open row actions on any league → click **Manage Hosts**
3. Verify owner shown with Crown badge (no remove button)
4. Search a user by email → click **Add** → toast confirms → user appears as co-host
5. Click **X** on a co-host → toast confirms → row disappears
6. Try adding the same user again → toast "Already a host"
7. Verify a non-admin calling any `/co-host` endpoint gets 403
8. Verify `DELETE` on `created_by` returns 409

## Checklist


blackhathckr (changes requested):
## Review — Admin co-host assignment (#526)

Clean feature structure — dialog UI, API routes, shared helper. However two blockers need fixing.

### Blockers

**1. Missing `leaguemembers` insertion when assigning co-host**
The POST handler calls `assignRoleToUser()` which only inserts into `assignedrolesforleague`. It does NOT add the user to `leaguemembers`. Compare with the impersonate route (`/api/admin/leagues/[leagueId]/impersonate/route.ts`, lines 60-79) which explicitly inserts into `leaguemembers` first, then assigns the role.

Without a `leaguemembers` row, the co-host won't appear as a league member — breaking navigation, league access checks, and queries that join on `leaguemembers`.

**Fix:** Before calling `assignRoleToUser()`, check if user exists in `leaguemembers` for that league. If not, insert them.

**2. Destructive role stripping without atomicity**
The POST handler removes ALL existing roles before assigning host:
```ts
for (const role of existingRoles) {
  await removeRoleFromUser(user_id, leagueId, role);
}
```
If `assignRoleToUser` fails after roles are removed, the user is left with zero roles. Also, silently stripping captain/governor roles could break team management.

**Fix:** Either use a transaction, or assign the new role first and only remove old roles after confirming success.

### High Priority

**3. No league existence check in POST handler**
GET and DELETE both validate the league exists. POST does not — invalid `leagueId` gives a cryptic FK error instead of a clean 404.

**4. `getHostRoleId()` null not checked in GET handler**
If the role lookup returns null, the query silently returns empty results with no error surfaced.

### Medium

- Search results don't filter out existing hosts — clicking "Add" on an existing host shows "Already a host" toast. Better UX: filter or disable them in results.
- `getHostRoleId` helper only used in GET — POST/DELETE use internal lookups via `assignRoleToUser`/`removeRoleFromUser`. Consider removing the helper or using it consistently.
- `requireAdmin` helper defined locally in GET/POST route but not reused in DELETE route — either extract to shared util or use inline consistently.

### Nits
- `HostUser` interface has optional `username?`/`email?` — UI renders `{h.username}` which could show nothing. Add fallback text.
- `results.slice(0, 8)` hardcoded — should be server-side limit.
- Dialog state resets on `open` change but doesn't account for `leagueId` changes.

blackhathckr (changes requested):
## Re-review: PR #530 — Multi-host admin panel

All blockers and high-priority issues from the first review are properly fixed:
- leaguemembers insertion on co-host assignment ✓
- Atomic role ordering (assign first, remove after) ✓
- League existence check ✓
- getHostRoleId null guard ✓
- Existing hosts filtered from search ✓
- requireAdmin extracted to shared util ✓
- UI fallback text for optional fields ✓

One nit: the sequential \`removeRoleFromUser\` loop doesn't check return values — if a removal fails silently, the user keeps both old role + host role. Acceptable since extra roles are less destructive than zero roles, but worth logging failures.

**Note:** This PR depends on mfl#525 (PR #543) for the underlying host role API. PR #543 currently has changes requested — once those are resolved and merged, this PR should work correctly on top of it.

Overall the code is solid. Just one remaining nit — the hardcoded \`.slice(0, 8)\` for search results. Not blocking but should be a server-side limit.

Fix the \`.slice(0, 8)\` to use a constant or move the limit server-side, then this is good to go.

blackhathckr (approved):
## Re-review #2: PR #530 — Admin co-host assignment

The remaining nit from the previous review is fixed:

**`.slice(0, 8)` hardcoded** — ✅ Replaced with `MAX_HOST_SEARCH_RESULTS` constant on client, `limit` query param passed to server, server applies `.limit()` with validation (`Number.isFinite && > 0`).

Also verified:
- `requireAdmin` extracted to shared `@/lib/auth/require-admin.ts`
- `HostUser` has proper fallback text (`username ?? email ?? 'Unknown user'`)
- Existing hosts filtered from search results
- League existence check in POST handler
- `getHostRoleId` null guard
- Atomic role ordering (assign first, remove after) with error logging

Clean implementation. Blocked by #543 — once that merges, this is ready.

PR #531, opened 2026-07-01
fix: SSR role resolution treating Supabase FK object as array

blackhathckr:
## Summary
- `getUserLeaguesWithRoles` in `league-query-service.ts` accessed `roles?.[0]?.role_name` but Supabase returns FK joins as an object (`{ role_name: "captain" }`), not an array
- This caused all role assignments to resolve as `undefined`, defaulting every user to `['player']` on initial SSR load
- Aligned with the working `/api/user/leagues` route which correctly uses `roles?.role_name`

## Impact
- Users with captain/governor/host roles assigned via `assignedrolesforleague` would only see "Player" on first page load
- Role switcher and role-specific sidebar sections (Captain Actions, Host Actions) would not appear until a client-side refetch was triggered

## Test plan
- [ ] Log in as a user with captain role in a league (e.g. mobilisation/scheduled phase)
- [ ] Verify captain sidebar and role switcher appear immediately on first load
- [ ] Verify host/governor roles also resolve correctly on first load


adminmfl (approved):
Verified fix — Supabase FK returns object not array.

PR #532, opened 2026-07-01
feat: implement forced password change workflow for new users and add required database migration

dhjr:
## Summary
This PR implements a mandatory first-login password change flow for pre-registered users. Users flagged with `needs_password_change = true` in the database are intercepted after login and forced to set a new password before accessing the application.

## Related Issue
Closes #523 

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] UI/UX improvement
- [ ] Documentation

## Changes Made
- Added a new database migration (`20260702000000_first_login_password_change.sql`) to introduce the `needs_password_change` column on the `users` table.
- Extended NextAuth types (`src/types/next-auth.d.ts`) and configuration (`src/lib/auth/config.ts`) to propagate the new flag into the active session. OAuth logins automatically clear/bypass this flag.
- Updated `src/app/(app)/layout.tsx` to intercept authenticated users who need a password change and redirect them to the new `/set-password` route.
- Created `src/app/set-password/page.tsx` as a focused UI for setting a new password without requiring the old one.
- Modified the update password API (`src/app/api/auth/update-password/route.ts`) to securely bypass the "current password" check if the user is flagged, and to clear the flag upon success.
- Updated the mobile login endpoint (`src/app/api/auth/mobile/login/route.ts`) to include the `needsPasswordChange` flag for mobile clients to handle redirection.

## How to Test
1. Apply the new Supabase migration locally to add the `needs_password_change` column.
2. Manually update a test user in your local database: `UPDATE public.users SET needs_password_change = true WHERE email = 'YOUR_TEST_EMAIL@example.com';`.
3. Log into the web application using the test user's email and password.
4. Verify that you are immediately redirected to `/set-password` and cannot access `/dashboard` or `/leagues`.
5. Submit a new password. Verify that you are redirected to `/complete-profile` (or the dashboard) and that the database flag is now `false`.
6. (Optional) Test the mobile login API endpoint via curl/Postman to ensure `needsPasswordChange: true` is returned for flagged users.

## Screenshots

image

## Checklist


blackhathckr (changes requested):
## Review — Forced password change workflow (#523)

Solid architecture — flag propagation through JWT/session is correct, server-side check in update-password reads the DB flag (not trusting client claims), OAuth users handled. Three blockers need fixing.

### Blockers

**1. No server-side password length validation**
`update-password` API only checks `if (!newPassword)` — accepts a 1-character password. Every other password endpoint uses `MIN_PASSWORD_LENGTH` from `@/types/auth`. The client-side `minLength={6}` is trivially bypassed.

**Fix:**
```ts
import { MIN_PASSWORD_LENGTH } from '@/types/auth';
if (newPassword.length < MIN_PASSWORD_LENGTH) {
  return NextResponse.json({ error: \`Password must be at least \${MIN_PASSWORD_LENGTH} characters\` }, { status: 400 });
}
```

**2. Admin layout bypass — no `needsPasswordChange` guard**
`src/app/(admin)/admin/layout.tsx` checks auth + admin role but NOT `needsPasswordChange`. An admin user flagged for password change can navigate to `/admin/*` and bypass the forced flow entirely. The redirect only exists in `(app)/layout.tsx`.

**Fix:** Add the same `needsPasswordChange` check + redirect to `/set-password` in the admin layout.

**3. `/complete-profile` bypass**
`src/app/complete-profile/page.tsx` is outside the `(app)` route group and has no `needsPasswordChange` check. A flagged user can navigate directly to `/complete-profile` to skip the password change.

**Fix:** Add `needsPasswordChange` redirect before the profile completion check.

### High Priority

**4. Hardcoded redirect to `/complete-profile` on success**
`set-password/page.tsx` always redirects to `/complete-profile` after password change. If the user already has a complete profile, they'll hit `/complete-profile` which immediately redirects them elsewhere. Use `window.location.href = '/'` and let existing session routing decide, or check `needsProfileCompletion` first.

**5. `needsPasswordChange` missing from `User` interface in `next-auth.d.ts`**
The `config.ts` augmentation adds it, but `next-auth.d.ts` User interface does not include it. Both files augment the same module — add it to `next-auth.d.ts` for consistency.

### Medium
- OAuth login silently clears `needs_password_change` in DB — confirm this is intentional product behavior (pre-registered user logs in via Google, bypasses password requirement)
- `(user as any)` casts in mobile login route — type the query result properly
- `.gitignore` adding `.agents` is out of scope — remove from this PR

### Nits
- Success alert uses hardcoded green colors — use theme-aware classes for dark mode
- Password toggle buttons missing `aria-label`
- Formatting-only changes (semicolons) inflate the diff — note in PR description or separate commit

blackhathckr (changes requested):
## Re-review: PR #532 — Pre-reg auth / forced password change

All 3 blockers and both high-priority items from the previous review are fixed. Nice work — the architecture is sound: `needs_password_change` flows correctly through DB → JWT → session, guards are in all 3 layouts (app, admin, complete-profile), and server-side validation now uses `MIN_PASSWORD_LENGTH`.

Also good: `.gitignore` reverted, `DbUser` type eliminates scattered `(user as any)` in mobile login.

### Remaining issues:

**1. Client-side hardcodes `6` instead of `MIN_PASSWORD_LENGTH` (medium)**
`src/app/set-password/page.tsx` — the client validation uses `password.length < 6`, `minLength={6}`, and placeholder "At least 6 characters". The server correctly imports `MIN_PASSWORD_LENGTH` from `@/types/auth`. Use the same constant on the client for consistency — if it ever changes, client and server will be out of sync.

**2. Unnecessary `(session as any)` casts (medium)**
You added `needsPasswordChange` to the `next-auth.d.ts` type declarations, but then cast through `(session as any).user?.needsPasswordChange` in:
- `src/app/(admin)/admin/layout.tsx` (2 occurrences)
- `src/app/complete-profile/page.tsx` (1 occurrence)
- `src/app/set-password/page.tsx` (2 occurrences)

Just use `session?.user?.needsPasswordChange` directly — the types already support it.

**3. `(user as any)` casts in `update-password/route.ts` (nit)**
The mobile login route was properly fixed with `DbUser`, but `update-password/route.ts` still uses `(user as any).needs_password_change` and `(user as any).password_hash`. Add a similar inline type.

**4. Success alert not dark-mode safe (nit)**
`set-password/page.tsx` — hardcoded `border-green-200 bg-green-50 text-green-800` won't adapt to dark mode.

blackhathckr (approved):
## Re-review #2: PR #532 — Forced password change

All 4 previously flagged issues are fixed:

1. **Client hardcodes `6`** — ✅ Both `set-password/page.tsx` and `complete-profile/page.tsx` now use `MIN_PASSWORD_LENGTH` from `@/types/auth` for validation, placeholder, and minLength
2. **Unnecessary `(session as any)` casts** — ✅ All 5 occurrences removed. Direct `session.user?.needsPasswordChange` works with the updated type declarations
3. **`(user as any)` in update-password** — ✅ Inline `DbUser` type replaces all `as any` casts
4. **Success alert not dark-mode safe** — ✅ Both pages add `dark:` classes for border, bg, and text colors

Additional fixes verified:
- `.gitignore` change removed from PR
- `DbUser` type in mobile login eliminates all `(user as any)` casts
- `needsPasswordChange` guard added to admin layout and complete-profile
- After password change, redirects to `/` (lets session routing decide)

Nit: Password toggle buttons in `set-password/page.tsx` (new file) are missing `aria-label` for screen readers. Non-blocking — can be addressed in a follow-up.

blackhathckr (changes requested):
## Retracting approval — 1 remaining nit

**Missing `aria-label` on password toggle buttons in `set-password/page.tsx`**

Both show/hide password toggle buttons (password + confirm) are icon-only `<Button>` elements with no `aria-label`. Screen readers announce them as unlabeled buttons.

**Fix:**
```tsx
<Button
  type="button"
  variant="ghost"
  size="icon"
  aria-label={showPassword ? 'Hide password' : 'Show password'}
  ...
>
```

Same for the confirm password toggle (`showConfirmPassword`).

Two-line addition. Fix this, then good to go.

blackhathckr (changes requested):
## Re-review of latest commit `6920e8e6`

**Fixed**: MIN_PASSWORD_LENGTH usage, removed `as any` casts, dark-mode safe success alert. All good.

**Still not fixed (1 nit)**: `aria-label` on password toggle buttons.

Both `set-password/page.tsx` and `complete-profile/page.tsx` have icon-only password toggle `<Button>` elements with no `aria-label`. Screen readers cannot identify what these buttons do.

**Fix** — add to each password toggle button:
```tsx
aria-label={showPassword ? 'Hide password' : 'Show password'}
```

Applies to:
- `set-password/page.tsx`: password toggle + confirm password toggle (2 buttons)
- `complete-profile/page.tsx`: password toggle (if present in this PR's diff)

blackhathckr (approved):
All nits resolved. aria-label added to all password toggle buttons in set-password and complete-profile pages.

PR #558, opened 2026-07-04
Fix/silent submission overwrite

dhjr:
## Summary
This PR fixes a bug where users could silently overwrite their existing daily submissions when a league has a strict daily submission limit (e.g., `max_submissions_per_day = 1`). It enforces daily limits correctly both on the frontend (button disable & form validation) and the backend (throwing an explicit error instead of a silent upsert).

## Related Issue
Closes #533 

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] UI/UX improvement
- [ ] Documentation

## Changes Made
- **`entries-mutations.ts`**: Updated `submitWorkout` to fetch the league's `max_submissions_per_day` and explicitly throw an error if the daily entry count meets or exceeds the limit, preventing silent overwrites.
- **`activity-form.tsx`**: Updated the submit button's `disabled` state to check `(isDailyMultiFrequency && !canSubmitDailyMulti)` so that it properly greys out for multi-frequency activities when their limits are hit, without accidentally blocking standard 1-per-day activities.
- **`use-submit-handlers.ts`**: 
  - Added an early return check in `handleValidationAndConfirm` to surface a toast error if a user exceeds the `isDailyMultiFrequency` limit before they attempt an API call.
  - Fixed a TypeScript error where `proof_url` and `proof_url_2` were passing `null` to properties strictly typed as `string | undefined` by adding `?? undefined`.

## How to Test
1. Set up a league with a daily submission limit of 1 (`max_submissions_per_day = 1`).
2. Log an activity (e.g., Running) for today.
3. Attempt to log a second activity of the same type for today. The UI should appropriately block you or warn you, and the backend should reject it rather than silently updating the previous entry.
4. Try logging an activity that explicitly allows multiple daily submissions (e.g., a custom 'Hydration' activity with `frequency = 3`). Verify that the button is only disabled once you have hit the 3rd submission for the day.

## Checklist


blackhathckr (dismissed):
Looks good — solid fix for the silent overwrite bug. Frontend and backend both enforce the daily limit correctly, and rejected entries are properly excluded from the count.

**Minor nits (non-blocking):**
1. `(member.data.leagues as any)?.max_submissions_per_day` — consider typing the join result instead of `as any`
2. The `overwrite` param passed from the frontend is now effectively dead code since duplicates always throw — can be cleaned up in a follow-up

blackhathckr (changes requested):
Good fix for the silent overwrite bug. Frontend + backend enforcement is correct, rejected entries properly excluded from count. Two nits to fix before approval:

1. **`entries-mutations.ts`**: `(member.data.leagues as any)?.max_submissions_per_day` — type the join result instead of using `as any`
2. **`use-submit-handlers.ts`**: The `overwrite` param is now dead code since duplicates always throw instead of updating — remove or handle it properly

blackhathckr (approved):
Both nits addressed:
1. Join result properly typed instead of `as any`
2. Overwrite logic fully removed across all 5 files — clean cleanup

Good to merge.

PR #559, opened 2026-07-05
feat: add maximum value constraints to submission reupload schema fields

dhjr:
## Summary
This PR adds `.max()` upper bounds to the measurement fields in the submission reupload API's Zod schema. This prevents users from submitting absurdly large values (e.g., 999,999,999 steps) and aligns the reupload validation limits with the existing manual entry API.

## Related Issue
Closes #535 

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] UI/UX improvement
- [ ] Documentation

## Changes Made
- Added `.max(1440)` to the `duration` field (24 hours).
- Added `.max(1000)` to the `distance` field (1,000 km).
- Added `.max(500000)` to the `steps` field (500,000 steps).
- Added `.max(36)` to the `holes` field (36 holes).
- Applied these bounds within `src/app/api/submissions/[id]/reupload/route.ts` to ensure the server rejects structurally invalid outliers before attempting to process the update.

## How to Test
1. Log in to the application and navigate to an activity that was previously rejected.
2. Attempt to reupload the submission using intercepting proxy tools (or raw cURL) to submit a payload containing metrics exceeding the bounds (e.g., `"steps": 9999999`).
3. Verify that the API immediately rejects the request with a `400 Bad Request` and a Zod validation error, rather than proceeding with processing.

## Checklist


blackhathckr (changes requested):
Max bounds on reupload schema look reasonable (1440 min, 1000 km, 500k steps, 36 holes).

However, this only adds bounds to the **reupload** API schema (`/api/submissions/[id]/reupload`). Does the main submission API (`/api/entries/upsert`) also need these max bounds? Issue #535 says "No max bounds on submission API" which likely covers both endpoints. Verify and add to the main upsert schema too if missing.

blackhathckr (approved):
## Re-review — PR #559

Verified — the main upsert route (`/api/entries/upsert`) already has the same max bounds:
- Duration: 1440 min (line 378)
- Distance: 1000 km (line 386)
- Steps: 500,000 (line 433)

The reupload schema in this PR mirrors those same constraints. LGTM.

PR #588, opened 2026-07-08
hotfix: SSR cache was bypassing team-size normalization

blackhathckr:
## Problem
Leaderboard SSR `initialData` was seeded into client cache. When the hook ran its first fetch, it hit the cache and returned unnormalized points — normalization pipeline never executed.

Fit Warriors (10 members) showed 20 points instead of 22 (normalized: 20 × 11/10).

## Fix
Removed the `initialData` cache seeding. SSR data still renders instantly via `useState` initializer, but the client fetch runs the full normalization pipeline and replaces it with correctly normalized data.

## Verified
Tested locally against prod DB — Fit Warriors now shows 22 points with 20 logs.


adminmfl (approved):
Verified locally. LGTM.

PR #599, opened 2026-07-09
fix: 7 dashboard UI polish fixes

blackhathckr:
## Summary
- Auto-upgrade phase to `league_active` when `start_date` is past (fixes "Mobilisation" showing on running leagues)
- Hide "Invite Members" card after league has started
- Remove TiebreakerInfo section from leaderboard
- Deduplicate rejected entries in Validate view when approved/pending replacement exists on same member+date
- Simplify "Avg RR — You vs Team vs League" label to "Avg RR"
- Change StatCell layout to inline (label left, value right) for Team Summary
- Clamp weekly navigation to league start date on dashboard timeline and manual-entry page

## Test plan
- [x] Phase indicator hidden on running leagues (FFL2)
- [x] Invite Members card not visible after league start
- [x] Tiebreaker text removed from leaderboard bottom
- [x] Rejected entries hidden in Validate when replacement exists
- [x] RR chart label shows "Avg RR" only
- [x] Team Summary stats show label+value on same line
- [x] Cannot navigate to weeks before league start on dashboard timeline or manual-entry


adminmfl (approved):
Verified all 7 fixes.

PR #601, opened 2026-07-09
fix: rest day overwrite loop + cron to 2pm IST

blackhathckr:
## Summary
- Fix infinite 409 loop when replacing a manually-submitted rest day with a workout
  - Root cause: upsert API's rest-day exclusivity check ignored the `overwrite` flag
  - Also: overwrite candidate filter only matched workout entries, missing the rest entry during cross-type overwrite
- Shift auto-rest-day cron from 4pm IST (10:30 UTC) to 2pm IST (08:30 UTC) per client request

## Test plan
- [x] Mark rest day → submit workout on same day → confirm overwrite → should succeed (no loop)
- [x] Auto-rest cron schedule updated in vercel.json


adminmfl (approved):
Verified fix.

adminmfl (approved):
Verified.

PR #606, opened 2026-07-10
fix: manual-entry RR calculation for distance-based activities

blackhathckr:
## Summary
- Fixed manual-entry `calculateRunRate` function that was branching by `measurementType` and only computing RR for the matching metric type
- Now computes all metrics (duration, distance, steps, holes) independently and returns the maximum, matching the shared `rr-calculator.ts` behavior
- Fixes issue where distance-based activities (e.g. Walking with 4.2km) showed valid RR on frontend but backend rejected with "RR must be at least 1.0"

## Test plan
- [x] Tested Walking + 4.2km distance in manual entry → RR 1.05 accepted (was previously rejected)
- [ ] Verify other activity types (duration-based, steps-based) still calculate correctly


adminmfl (approved):
Fix verified on prod test league — Walking 4.2km now correctly accepted with RR 1.05

PR #607, opened 2026-07-11
fix: age-group overrides for steps minimum and RR calculation

blackhathckr:
## Summary
- Fixed age_group_overrides being skipped when min_value is null in upsert, preview-rr, and manual-entry backends
- Fixed submit page frontend RR estimate not passing userAge to getConfiguredMinimum
- Fixed manual-entry page to resolve selected member's age for correct RR preview and minimum label
- Affects both player submit page and host manual entry page

## Test plan
- [x] Verified Sushil (age 66) can submit 7,020 steps with correct RR ~1.40 on submit page
- [x] Verified manual entry shows correct minimum (5,000) and RR for age-adjusted members
- [x] Verified no regression for activities with min_value set


adminmfl (approved):
Verified — age overrides now work independently of min_value

adminmfl (approved):
Verified — age overrides now work independently of min_value

PR #608, opened 2026-07-11
fix: Avg RR chart — one-line layout, bigger dots, highlighted numbers

blackhathckr:
## Summary
- All 3 values (You / Team / League) now rendered on a **single line** using `justify-between` — "League" no longer wraps to a second row on mobile
- Legend dots increased from `w-2` → `w-3`, chart marker dots from `w-2.5` → `w-3.5`
- RR numbers now highlighted in their **matching dot colors** (You=red, Team=primary, League=amber) instead of generic foreground color

## Test plan
- [ ] Open My Zone on mobile viewport (375px width)
- [ ] Verify all 3 values (You / Team / League) appear on **one line**
- [ ] Verify dots are visibly larger than before
- [ ] Verify numbers are colored: You in red, Team in green/primary, League in amber/orange
- [ ] Verify chart marker dots on the scale bar are also larger


adminmfl (approved):
LGTM

PR #621, opened 2026-07-11
fix: default all submissions to approved status

blackhathckr:
## Summary
- All submissions now default to `approved` status — no more `pending`
- Captains/governors/hosts can still reject entries after the fact
- Plausibility flagging still sets `review_tier` for audit routing but no longer blocks with pending status
- Changed in 4 locations: main upsert flow, plausibility check, reupload flow, legacy mutations path

## Test plan
- [ ] Submit a workout as a regular player — verify status is `approved` immediately
- [ ] Submit a workout with suspicious values (very high distance) — verify status is still `approved` but `review_tier` is set
- [ ] Reupload a rejected submission — verify status is `approved`
- [ ] Verify leaderboard counts the entry immediately


adminmfl (approved):
LGTM — all submissions default to approved

adminmfl (approved):
LGTM — all submissions default to approved

PR #622, opened 2026-07-11
fix: auto-rest dedup bug and captain validate view

blackhathckr:
## Summary

- **Root cause fix**: Auto-rest entries were not deleted when a player/host submitted a workout for the same date. The deletion check was placed *after* workout_type filtering, which excluded rest entries (they have `workout_type=null`). Moved the check to *before* filtering so it correctly finds and deletes the auto-rest.

- **Captain validate view**: Added dedup logic to `my-team/submissions` API — was completely missing. Captains previously saw both rejected auto-rest and the replacement workout as separate entries.

- **Host validate view**: Enhanced existing dedup to also hide approved rest entries when an approved workout exists for the same member+date (handles stale auto-rest that wasn't cleaned up).

## Test plan

- [ ] Submit a workout for a date that has an auto-rest entry → auto-rest should be deleted, only workout remains
- [ ] Captain validate view should not show rejected entries when approved exists for same member+date
- [ ] Host validate view should not show duplicate approved rest+workout for same member+date
- [ ] Overwrite flow (host with overwrite=true) still works correctly
- [ ] Normal submission without auto-rest continues to work as before


adminmfl (approved):
LGTM — fixes root cause of auto-rest not being deleted when workout submitted (workout_type filtering excluded rest entries), adds missing captain view dedup, and enhances host view dedup for stale approved rest entries.

PR #624, opened 2026-07-11
feat: overlapping challenge date warning with override (#550)

GitGuru-sudo:
Summary

  Block challenges whose dates overlap an active/scheduled challenge in the same league; host can acknowledge the
  warning and override on create or edit.

  Related Issue

  Closes #550

  Type of Change

  - [ ] Bug fix
  - [x] New feature
  - [ ] UI/UX improvement
  - [ ] Documentation

  Changes Made

  - datesOverlap() util + unit tests for all overlap combinations (partial, full, containment, shared endpoint, null
  bounds)
  - Overlap gate (409 CHALLENGE_OVERLAP) on POST and PATCH challenge routes; skipped when overrideOverlap is set
  - OverlapWarningDialog surfaces conflicting challenges and an override action, wired into create/edit/activate/finish
  flows
  - Record host overrides in new challenge_overlap_acknowledgements table

  How to Test

  1. Create a challenge with dates in a league so it's active/scheduled.
  2. Create or edit another challenge with dates that overlap it → warning dialog lists the conflict and blocks by
  default.
  3. Click "Override & continue" → challenge saves and the override is recorded in challenge_overlap_acknowledgements.
  4. Repeat via date editing on an existing challenge to confirm the gate fires on update too.

  Screenshots / Recordings

  Checklist


blackhathckr (dismissed):
## Review: Overlapping Challenge Date Warning (#550)

Clean implementation. The architecture is solid:

- **`datesOverlap` utility** — correct interval overlap formula (`aStart <= bEnd && bStart <= aEnd`), handles null bounds. 8 unit tests covering all overlap combinations.

- **409 `CHALLENGE_OVERLAP` pattern** — API returns structured error with conflict list, client intercepts and shows dialog. `submitWithOverlapGuard` wrapper is DRY — one function wraps all 4 submission paths (create, activate, edit, finish).

- **Override audit trail** — `challenge_overlap_acknowledgements` table with FKs, RLS enabled, service-role only writes. Good compliance pattern.

- **`NON_BLOCKING_STATUSES`** — draft/closed/published challenges don't trigger overlap. Correct.

### Minor nits (non-blocking)

- **`console.error` in `challenge-overlap.ts`** (L495): prefer `logger.error`.

- **Silent swallow on query error** (`findOverlappingChallenges` L468): `if (error || !data) return []` — if the query fails, the gate silently passes and the challenge is created without overlap checking. Consider logging the error at minimum.

- **`datesOverlap` shared-endpoint = overlap** — `aStart <= bEnd && bStart <= aEnd` means if challenge A ends on Jul 10 and challenge B starts on Jul 10, they overlap. This is correct since both challenges are active on that day. Just confirming this is intentional.

LGTM — approve.

blackhathckr (changes requested):
Dismissing previous approval — nits must be fixed before approval.

The architecture and implementation are solid (see previous review). Please fix these before re-requesting review:

1. **`console.error` in `challenge-overlap.ts`** (L495): use `logger.error` from `@/lib/logger` — project convention.

2. **Silent swallow on query error** in `findOverlappingChallenges` (L468): `if (error || !data) return []` — if the query fails, the overlap gate silently passes and the challenge is created without any overlap checking. At minimum log the error; ideally return an error so the caller can decide (e.g. block creation when the check itself fails).

blackhathckr (approved):
Both items fixed. Overlap check now fails closed on query error — correct behavior. Logger used throughout. LGTM.

PR #627, opened 2026-07-11
feat(challenges): host swap deployed challenge with library template …

GitGuru-sudo:
## Summary
  Add a Swap action to each deployed challenge in configure-challenges so a host can replace it in place with any
  Available/Default stage template, preserving the date slot and row id.

  ## Related Issue
  Closes #552

  ## Type of Change
  - [ ] Bug fix
  - [x] New feature
  - [ ] UI/UX improvement
  - [ ] Documentation

  ## Changes Made
  - New POST `/api/leagues/[id]/challenges/[challengeId]/swap` route: host or governor gated, refuses swaps on
  challenges with submissions, validates the target is available/default stage and not already deployed.
  - Auto-configure the swapped challenge from the template to match the activation/wizard path: weight_loss targets get
  default tiers and tier-derived total_points; other types clear stale weight-loss config.
  - Swap picker reuses availablePresets (Available + Default, deployed excluded); Swap button disabled when the
  challenge has submissions.

  ## How to Test
  1. As a host/governor, open a league's Configure Challenges page with a deployed default challenge.
  2. Click **Swap** on a challenge card and pick an Available/Default template from the picker.
  3. Confirm the challenge is replaced in place, the date slot is preserved, and (for weight_loss) tiers and points are
  auto-configured.
  4. Verify the Swap button is disabled on a challenge that already has submissions.

  ## Screenshots / Recordings

  ## Checklist


blackhathckr (changes requested):
## Review: Host Swap Challenge (#552)

Good feature — in-place swap preserving date slot and row ID. Submission guard, uniqueness guard, and lifecycle stage validation are all correct.

### Must fix

**1. BUG: total_points not reset for non-weight-loss swaps**

swap/route.ts L251-257:

```ts
if (targetTemplate.challenge_type === 'weight_loss') {
  updates.config = { ...DEFAULT_WEIGHT_LOSS_CONFIG };
  updates.total_points = maxTierPoints(DEFAULT_WEIGHT_LOSS_CONFIG.tiers);
} else {
  updates.config = null;
  // total_points is NOT set here — old value carries over
}
```

If a weight-loss challenge (1000 points) is swapped for a standard challenge (500 points), the old total_points (1000) carries over because only config is cleared. Add:

```ts
updates.total_points = targetTemplate.total_points ?? null;
```

for the non-weight-loss branch. This also covers standard-to-standard swaps where point values differ between templates.

### Nits (non-blocking)

- **console.error in API route** (lines 166, 214, 269, 285): prefer logger.error from @/lib/logger.

- **UI: current template not filtered from swap list** — if the deployed challenge uses template T1, the user can select T1 in the picker. Backend correctly rejects it (uniqueness guard returns 409), but the UX is confusing. Filter out the current challenge_id from the presets list on the client side.

- **IIFE pattern** (configure-challenges L50-70): the immediately-invoked function expression can be simplified by extracting the hasSubmissions check before the JSX return.

Fix the total_points bug, then this is ready.

blackhathckr (approved):
total_points reset fix is correct — uses target template points. Logger, IIFE cleanup, and self-swap filter are nice extras. LGTM.

PR #629, opened 2026-07-12
Feat/activity submission

dhjr:
## Summary
This PR introduces two major enhancements to the activity submission system. First, it allows league hosts to enforce a strict local-time deadline (e.g., 08:00 AM) for activity submissions. Second, it migrates daily submission limits from a global league setting to a per-activity configuration. This unlocks the ability for users to submit multiple different activity types (e.g., Meditation + Yoga) on the same day while still enforcing strict frequency limits for identical activities.

## Related Issue
Closes #483 , #484 

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] UI/UX improvement
- [ ] Documentation

## Changes Made
- Added `submission_cutoff_time` to the `leagueactivities` table and implemented timezone-aware validation using the user's local time during entry upsert.
- Dropped the restrictive global `min_submissions_per_day` and `max_submissions_per_day` limits from the `leagues` table (undoing the global limitations introduced in #466) via a SQL migration.
- Added a `min_frequency` column to `leagueactivities` to pair with the existing max `frequency` limit.
- Removed the global max submission block from `api/entries/upsert/route.ts` to natively support submitting different activity types on the same calendar day. 
- Updated the League Activity configuration UI (`ActivityMinimumDropdown`) to allow setting the submission cutoff time and specific Min/Max frequency requirements.
- Cleaned up deprecated global limit fields across all league forms and endpoints.

## How to Test
1. Run the new database migrations locally (`20260713_submission_cutoff.sql` and `20260713_per_activity_limits.sql`).
2. Go to League Settings -> Activities. Verify the old global min/max limits have been removed from the main UI.
3. Configure an activity (e.g., Meditation) with a Submission Cutoff Time of `08:00`. Attempt to log a Meditation after 8 AM in your local time and verify it is appropriately blocked with a helpful error.
4. Configure two activities (e.g., Yoga and Meditation). Go to the submission form and submit a Yoga session for today, then submit a Meditation session for the same day. Verify both are successfully accepted without prompting an overwrite.

## Checklist


blackhathckr (dismissed):
## Review: Activity Submission — Cutoff Time + Min Frequency (#629)

Good feature additions — submission cutoff time enforcement and min_frequency config. The `getUserLocalTimeHHMM` utility and HH:mm validation regex are clean. However, there are critical issues.

### CRITICAL: Runtime crash in frequency change handler

In the activities page, the new `useEffect` calls:
```ts
onFrequencyChange(activityId, freqVal, minFreqVal, frequencyTypeDraft);
```

where `freqVal` is `number | null`. But the parent handler `handleFrequencyChange(activityId: string, value: string)` calls `value.trim()` — which will crash at runtime because numbers don't have `.trim()`. When `value` is `null`, `null.trim()` also throws.

The old direct call `onFrequencyChange?.(activityId, nextValue)` in `onValueChange` was removed, so the only call path is this useEffect. This will crash the admin UI when configuring activity frequencies.

Fix: Either update the parent handler to accept `number | null`, or convert to string before calling: `onFrequencyChange(activityId, String(freqVal ?? ''), ...)`.

### Must-fix items

1. **`max_submissions_per_day` enforcement deleted (65 lines)** — the entire daily submission limit feature is removed. If this is intentional, document it in the PR description. If not, this is a major regression — leagues with daily limits will no longer enforce them.

2. **`(leagueRow as any).status`** and similar `as any` casts in `upsert/route.ts` — these are pre-existing, but since you're modifying this file, worth cleaning up. Import or define the league row type properly.

3. **`(updated as any)?.min_frequency`** and **`(updated as any)?.frequency_type`** — new `as any` casts on the update result in the activities PATCH handler. Type the Supabase update response properly.

4. **Cutoff time race condition** — the cutoff check uses the user's local time at request time. If the user's device clock is slightly behind and they submit at 23:59 but the server receives it after midnight, the check could behave unexpectedly. Consider documenting this limitation or using server time as fallback.

5. **`submission_cutoff_time` not in the GET response** — the activities GET handler now selects `min_frequency` but doesn't appear to select `submission_cutoff_time`. Verify the client can read cutoff time to display the hint ("Submissions must be logged before HH:mm").

blackhathckr (changes requested):
## Code Review -- PR #629: Feat/activity submission

**Files changed (18):**
1. `src/app/(app)/leagues/[id]/activities/page.tsx` -- adds `submission_cutoff_time` to pending changes map
2. `src/app/(app)/leagues/[id]/submit/_components/activity-form/activity-form.tsx` -- passes cutoff time to DateSelector
3. `src/app/(app)/leagues/[id]/submit/_components/activity-form/date-selector.tsx` -- shows cutoff time hint in UI
4. `src/app/api/entries/upsert/route.ts` -- enforces cutoff time; deletes global `max_submissions_per_day` block
5. `src/app/api/leagues/[id]/activities/route.ts` -- PATCH/GET support for `submission_cutoff_time` + `min_frequency`
6. `src/app/api/leagues/[id]/route.ts` -- removes `min/max_submissions_per_day` Zod refinement
7. `src/components/league/activities/activity-toggle-card.tsx` -- passes new fields through
8. `src/components/leagues/activity-minimum-dropdown.tsx` -- UI for cutoff time + min frequency
9. `src/components/leagues/settings/hooks/use-league-settings-form.ts` -- replaces global limits with `challenge_frequency_days`
10. `src/components/leagues/settings/sections/activities-section.tsx` -- removes global min/max submissions UI card
11. `src/hooks/use-league-activities.ts` -- `updateFrequency` now accepts `minFrequency`
12. `src/lib/services/leagues/league-mutation-service.ts` -- replaces global limit fields with `bonding_automations_enabled`
13. `src/lib/services/leagues/league-query-service.ts` -- removes `min/max_submissions_per_day` from SELECT
14. `src/lib/services/leagues/league-types.ts` -- removes `min/max_submissions_per_day` from `DbLeagueRow`
15. `src/lib/utils/timezone.ts` -- adds `getUserLocalTimeHHMM()` + formatting cleanup
16. `src/types/league-minimums.ts` -- adds `submission_cutoff_time` to `LeagueActivityWithMinimums`
17. `supabase/migrations/20260713_per_activity_limits.sql` -- adds `min_frequency`, drops global limit columns
18. `supabase/migrations/20260713_submission_cutoff.sql` -- adds `submission_cutoff_time` column

---

### Critical Issues

#### 1. `onFrequencyChange` type mismatch -- runtime crash

The new `useEffect` in `activity-minimum-dropdown.tsx` calls:
```ts
onFrequencyChange(activityId, freqVal, minFreqVal, frequencyTypeDraft);
```
where `freqVal` is `number | null` and `minFreqVal` is `number | null`.

But the prop type is still:
```ts
onFrequencyChange?: (activityId: string, frequency: string) => void;
```

And the parent handler in `page.tsx` is:
```ts
const handleFrequencyChange = (activityId: string, value: string) => {
    ...
    const trimmed = value.trim(); // CRASH: number.trim() is not a function
```

This will throw a **runtime TypeError** as soon as any frequency value changes in the dropdown. The `minFrequency` and `frequencyTypeDraft` args are silently dropped and never saved to `pendingChanges`.

**Fix:** Update the `onFrequencyChange` prop type, the `handleFrequencyChange` handler in `page.tsx`, and the `activity-toggle-card.tsx` interface to match the new 4-arg signature. The handler needs to write both `frequency`, `min_frequency`, and `frequency_type` into `pendingChanges`.

#### 2. Cutoff time blocks backdated submissions incorrectly

In `entries/upsert/route.ts`, the cutoff check compares the user's **current local time** against the cutoff:
```ts
if (localTimeHHMM > cutoffTime) {
    return NextResponse.json({ error: ... }, { status: 400 });
}
```

This runs unconditionally regardless of the submission date. If the cutoff is 08:00 and a user submits at 09:00 for **yesterday**, the submission is blocked even though the cutoff should only apply to today's date.

**Fix:** Add a date check -- only enforce the cutoff when the submission's `normalizedDate` equals the user's current local date. Something like:
```ts
const userLocalDate = getUserLocalDateYMD({ ... });
if (normalizedDate === userLocalDate && localTimeHHMM > cutoffTime) { ... }
```

---

### Significant Issues

#### 3. No `min_frequency <= frequency` validation in PATCH endpoint

The PATCH endpoint in `leagues/[id]/activities/route.ts` accepts both `frequency` (max) and `min_frequency` (min) but never validates that `min_frequency <= frequency`. A host could set min=5, max=3 which is contradictory. Add server-side validation.

#### 4. `min_frequency` is stored but never enforced

`min_frequency` is persisted to `leagueactivities` and exposed in the GET API, but the `entries/upsert` route never checks it. If this is intentional (e.g., enforcement comes in a follow-up PR), add a comment or TODO. Otherwise, it is dead config that misleads hosts into thinking there is a minimum.

#### 5. Deleted `max_submissions_per_day` enforcement -- verify no regressions

The PR removes ~65 lines of global max submission enforcement. The PR description says this is intentional to support submitting different activity types on the same calendar day. This is a behavioral change -- previously, if `max_submissions_per_day` was 1, submitting Yoga + Meditation on the same day was blocked. Now it is allowed.

Confirm: is there still per-activity-type duplicate detection downstream? The frequency-based duplicate check further down the route should handle same-type duplicates, but the cross-type daily cap is now gone entirely. Make sure this is the intended behavior and not just removing the guard without a replacement.

---

### Minor Issues

#### 6. `(updated as any)?.min_frequency` -- new `as any` cast

In `leagues/[id]/activities/route.ts`:
```ts
min_frequency: hasMinFrequency
    ? (min_frequency ?? null)
    : ((updated as any)?.min_frequency ?? null),
```

I know the `(updated as any)` pattern is pre-existing throughout this file (for `frequency`, `frequency_type`, etc.), but each new field added with `as any` compounds the tech debt. Consider typing the Supabase response properly.

#### 7. Unrelated change: `bonding_automations_enabled` in `league-mutation-service.ts`

The PR replaces `min_submissions_per_day`/`max_submissions_per_day` with `bonding_automations_enabled` in the mutation service. While removing the old fields makes sense, adding `bonding_automations_enabled` handling is unrelated to the activity submission feature. Was this intentional, or should it be in a separate PR?

#### 8. `useEffect` dependency completeness

In `activity-minimum-dropdown.tsx`, the new useEffect that clamps frequency values:
```ts
useEffect(() => {
    if (frequencyDraft !== '' && Number(frequencyDraft) > maxAllowed) { ... }
    if (minFrequencyDraft !== '' && Number(minFrequencyDraft) > maxAllowed) { ... }
}, [frequencyTypeDraft, frequencyDraft, minFrequencyDraft]);
```
references `maxAllowed` (derived from `frequencyTypeDraft`) but does not include it in the dependency array. It works because `maxAllowed` is derived from a dep that IS listed, but the linter would flag this. Either inline the computation or add `maxAllowed` to deps.

---

### What looks good

- `getUserLocalTimeHHMM()` implementation is solid -- handles IANA timezone, offset fallback, and legacy offset correctly
- Cutoff time validation regex in the PATCH endpoint is correct
- `.slice(0, 5)` to handle `time` column returning `HH:mm:ss` is a good defensive measure
- Migration files are clean and independent
- Removal of global submission limits from the settings UI, form hook, Zod schema, query service, and types is thorough and consistent
- The DateSelector cutoff hint with Clock icon is a nice UX touch

**Verdict:** Issues #1 (runtime crash) and #2 (backdated submissions blocked) need to be fixed before merge.

blackhathckr (changes requested):
Good fixes on the critical bugs -- cutoff now scoped to current day, handleFrequencyChange properly typed, min_frequency validation added.

Remaining issue: **5x `as any` in added code:**

In `src/app/(app)/leagues/[id]/activities/page.tsx`:
```ts
submission_cutoff_time: (activityFromMap as any).submission_cutoff_time,
```

In `src/hooks/use-league-activities.ts` (2 locations):
```ts
min_frequency: (la as any).min_frequency ?? null,
submission_cutoff_time: (la as any).submission_cutoff_time ?? null,
```

These fields exist on the `leagueactivities` response but the TypeScript type doesn't include them. Fix by extending the LeagueActivity interface (already has `submission_cutoff_time` in the API route interface) and updating the hook's type to match.

blackhathckr (approved):
Re-review of `2971c49` — the last remaining item (5× `as any` casts) is fully resolved:

- `use-league-activities.ts`: `LeagueActivity` now declares `min_frequency?: number | null` and `submission_cutoff_time?: string | null`, and the two casts on the hook's mapping are gone.
- `activities/page.tsx`: the local activity type declares both fields; `activityFromMap` accesses them without a cast.
- `activities/route.ts`: a proper `LeagueActivityRow` interface replaces the untyped rows.

Verified:
- No new `as any` introduced by the fix commit.
- All three touched files typecheck at **0 errors**.
- Lint actually improves — `route.ts` drops from 47 → 23 warnings, `page.tsx` 7 → 6; the residual `: any` annotations are pre-existing baseline (develop has more). No new warnings.
- Merges cleanly into develop (new-style `git merge-tree`, zero conflicts).

Everything from the earlier rounds (cutoff scoping to current day, typed `handleFrequencyChange`, `min_frequency` validation) was already verified fixed. LGTM — approving.

PR #654, opened 2026-07-16
fix: co-host settings access + rename Manual Workout to Activity

blackhathckr:
## Summary
- Fix co-host users getting "Access Restricted" on league settings despite having host role
- Root cause: Supabase FK join `roles(role_name)` returns a single object for many-to-one relations, but code accessed it as an array with `roles?.[0]?.role_name` — always `undefined`
- Fixed in `league-member-service.ts` and both admin memberships routes
- Rename "Manual Workout Entry" → "Manual Activity Entry" in sidebar and page heading

## Files changed
- `src/lib/services/leagues/league-member-service.ts`
- `src/app/api/admin/leagues/[leagueId]/memberships/route.ts`
- `src/app/api/admin/leagues/[leagueId]/memberships/[membershipId]/route.ts`
- `src/lib/navigation/sidebar-config.ts`
- `src/app/(app)/leagues/[id]/manual-entry/page.tsx`


adminmfl (approved):
Verified locally — co-host settings access fixed, label renamed.

PR #655, opened 2026-07-16
fix: delete rest day when host overwrites via manual entry

blackhathckr:
## Summary
- Manual entry endpoint only looked for existing entries matching the same type (workout), so rest day entries for the same date were never deleted
- Now explicitly deletes any rest day entries before inserting/updating the workout
- Fixes incorrect rest day total when host uses manual entry to overwrite a rest day

## Files changed
- `src/app/api/leagues/[id]/manual-entry/route.ts`


adminmfl (approved):
Verified.

PR #666, opened 2026-07-18
fix: dedup approved entries by date only, prefer workout over rest

blackhathckr:
## Summary
- Fixed dedup key in all 5 locations (dashboard-service.ts, dashboard.ts, report-rankings-calculator.ts) to use date-only key instead of date+workout_type
- Prevents rest and workout entries for the same member+date from both being counted, which inflated rest-day counts and point totals
- Prefers workout entries over rest entries when duplicates exist

## Test plan
- [ ] Verify leaderboard points match expected values for FFL league
- [ ] Verify rest day counts are accurate on player dashboards
- [ ] Check reports/rankings are consistent with dashboard


adminmfl (approved):
Dedup fix verified across all 5 locations.

PR #667, opened 2026-07-18
fix: rest day count derived from deduped entries

blackhathckr:
## Summary
- Rest day count was inflated when orphaned rest entries coexisted with workouts for the same date
- Changed all rest count calculations to derive from deduped entry lists that prefer workouts over rest
- Fixed in dashboard.ts, dashboard-service.ts, team-member-service.ts, ai-insights.ts, health-context-builder.ts

## Test plan
- [ ] Verify rest days used count matches actual unique rest dates on dashboards
- [ ] Verify AI insights rest day context is accurate
- [ ] Verify team member rest day counts on My Team page


adminmfl (approved):
Rest day count fix verified.

PR #672, opened 2026-07-18
hotfix: missing ArrowLeftRight import crashes configure-challenges page

blackhathckr:
## Summary
- `ArrowLeftRight` icon from lucide-react was used but never imported in configure-challenges page
- Causes ReferenceError that crashes the entire page in production
- One-line fix: add the missing import

## Test plan
- [ ] Verify configure-challenges page loads without error on production


adminmfl (approved):
Hotfix verified — missing import.

PR #673, opened 2026-07-18
fix: remove auto-balance button from team management

blackhathckr:
## Summary
- Removed auto-balance button from team settings page
- Client confirmed teams should not be rebalanced once formed
- Cleaned up unused handler, state, and imports (Scale, Loader2)

## Test plan
- [ ] Verify team management page loads without auto-balance button
- [ ] Verify balance score display still shows correctly
- [ ] Verify Create Team and other buttons still work


adminmfl (approved):
LGTM

PR #680, opened 2026-07-20
Fix/leaderboard empty placeholder

dhjr:
## Summary
This PR implements a motivational empty state across all leaderboard tables when a league has no activity (all scores are zero). It replaces the default ranked list (which misleadingly awarded medals for zero points) with a friendly call-to-action, while preserving the normal ranked list for mixed-score scenarios.

## Related Issue
Closes #668 

## Type of Change
- [ ] Bug fix
- [x] New feature
- [x] UI/UX improvement
- [ ] Documentation

## Changes Made
- Created a new `LeaderboardEmptyState` component to consistently display the motivational "no activity yet" message.
- Updated `calculateLeaderboard` in `src/lib/services/leaderboard-logic.ts` to calculate and return a global `allZero` boolean flag based on total league points.
- Extended the `LeaderboardData` interface in `src/hooks/use-league-leaderboard.ts` to support the `allZero` property.
- Refactored `league-individuals-table.tsx`, `league-teams-table.tsx`, `league-subteams-table.tsx`, and `realtime-scoreboard-table.tsx` to accept an `isAllZero` prop and conditionally render `LeaderboardEmptyState` instead of their respective tables.
- Updated `leaderboard-client-container.tsx` to pass the `allZero` flag down to child table components, preventing duplicated logic.

## How to Test
1. Start a new league (or use an existing league with no logged activity).
2. Navigate to the league's leaderboard page.
3. Verify that the "The race hasn't started yet!" empty state is displayed in place of the Teams and Top Performers tables.
4. Verify that the empty state also appears properly in the Sub-teams and Realtime Scoreboard components if enabled.
5. Check a challenge tab without any logged activity to ensure it independently renders the empty state.
6. Log a single workout for one player and verify that the normal ranked list correctly replaces the empty state on page refresh.

## Screenshots / Recordings
Screenshot from 2026-07-21 00-45-56

## Checklist


blackhathckr (changes requested):
Solid implementation — verified end to end. The `allZero: totalLeaguePoints === 0` flag is exactly right: it reuses the same condition that already drives the alphabetical rank-0 placeholder list in `leaderboard-logic.ts` (lines 607/731), so the empty state replaces precisely the "fake ranked list" the issue describes, and the mixed-score case is untouched. Verified:

- All rendered leaderboard surfaces covered: `LeagueTeamsTable` + `LeagueIndividualsTable` (via container prop) and `ChallengeSpecificLeaderboard` (independent per-challenge zero check — correct, since a challenge tab can be empty while the league isn't). `LeagueSubTeamsTable` / `RealTimeScoreboardTable` have no current call sites, so wiring the prop with a `false` default is harmless and future-proof.
- No hook-rule violations: the early returns in teams/subteams tables sit after all hook calls; individuals table has no hooks in its own body (they live in `RankBadge`).
- Removed `selectedChallenge` / `activityPts` / `challengePts` were verified dead on `develop`.

One nit to fix before approval:

**Nit: `as Record<string, unknown>` cast introduces a new TS error**

`challenge-specific-leaderboard.tsx` (~line 150):

```ts
visible.map((c) => (c as Record<string, unknown>).template_id || '')
```

`tsc` now reports TS2352 here ("neither type sufficiently overlaps") — this is a new error relative to `develop` (the old `as any` didn't produce one; it only builds because `ignoreBuildErrors` is on). Since the intent is that some challenge rows carry a `template_id`, add `template_id?: string | null` to the file's local challenge type and drop the cast entirely (`c.template_id ?? ''`). Cleaner than swapping one cast for another.

blackhathckr (approved):
Re-review of `0dedfe7` — the blocking nit is resolved. `template_id?: string | null` is now declared on the `Challenge` interface and the cast is gone (`c.template_id ?? ''`), which clears the TS2352 error. Verified `challenge-specific-leaderboard.tsx` now typechecks with 0 errors, and `leaderboard-logic.ts` holds at the same 22 pre-existing errors as develop (the one-line `allZero` addition introduces nothing new).

Also confirmed this branch still merges cleanly into develop despite the Logs-column removal (#692) landing on the same `league-teams-table.tsx` — the two edits are in different regions (new-style `git merge-tree` reports zero conflicts). All surfaces from the original review remain correctly covered. LGTM — approving.

PR #684, opened 2026-07-21
fix: promote My Zone custom activity name fix to main

blackhathckr:
Promotes #681 (custom activity names on My Zone timeline) from staging to main.


adminmfl (approved):
Approved for production release.

PR #679, opened 2026-07-20
refactor(challenges): decompose configure-challenges page, drop as an…

GitGuru-sudo:
## Summary
Module A of the #556 Phase 2 production-grade cleanup: decomposes the 1847-line configure-challenges page into child components under `components/challenges/configure/`, removes 17 `as any` casts and all `console.*` from the challenge submission/validate API routes. Also includes (and supersedes) PR #656: inverted challenge date-range validation (client + server) and the dashboard hydration fix.

## Related Issue
Part of #556 (Module A — Challenges & Configure). Closes #656.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [x] Refactor (no functional change)
- [ ] UI/UX improvement
- [ ] Chore (dependencies, config, CI)
- [ ] Documentation

## Changes Made
- Split `configure-challenges/page.tsx` (1847 lines) into presentational children: `challenge-card-grid`, `create-challenge-dialog`, `edit-challenge-dialog`, `review-dialog`, `finish-dialog`, `delete-alert`, `view-proof-dialog` + shared `types` — all state and handlers stay in the page, JSX moved verbatim.
- Removed 17 `as any` casts by typing Supabase result rows and swapped `console.*` → `logger` in the 3 challenge submissions/validate API routes; 2 `console.error` → `logger` in the page.
- Review-round fixes: restored the `presets.length === 0` empty-state condition; kept functional state updaters in ReviewDialog (stale-closure safety); dropped unused props (`loading`/`error` on ChallengeCardGrid, `selectedFile` on CreateChallengeDialog); removed the orphan `sub-team-list.tsx`.
- Intentional removal: the dead `setActivateOpen(false)` call in `handleOverrideConfirm` — it has no matching state declaration on `develop` (pre-existing dead/buggy line), so it is not carried over.
- From #656: end date can no longer be before start date — `min` attr on end-date pickers + guard toasts in create/edit/finish handlers + 400 responses in the challenges POST/PATCH routes; `StatsGrid` hydration mismatch fixed by gating role-dependent WhatsApp buttons behind mount.

## How to Test
1. `pnpm run dev`, log in as a host, open `/leagues/{id}/configure-challenges` (Console + Network tabs open).
2. Empty state: league with 0 challenges but presets available → "No challenges yet." must NOT show; with 0 challenges and 0 presets → it shows.
[...]

PR #685, opened 2026-07-21
refactor(activities): type league activities route rows, decompose dr…

GitGuru-sudo:
## Summary
Pure refactor — decomposes the 894-line `activity-minimum-dropdown.tsx` and 855-line activities page into focused components, types the route to remove 37 `as any` casts, and swaps raw `console.error` for the shared logger.

Part of  #556 

- [x] **Refactor (no functional change)**

**Changes Made:**
- Typed `LeagueActivityRow` in `lib/services/activities/` matching the GET `.select()` shape; removed all 37 `as any` in the route
- Replaced 9 `console.error` calls with structured `logger.error`
- Decomposed `activity-minimum-dropdown.tsx` (894→389 lines) into `MinimumRuleRow`, `MinimumRuleEditor`, `ActivityMinimumList`, `basic-settings`, `proof-fields`, `outcomes` under `components/leagues/activity-minimum/`
- Decomposed `activities/page.tsx` (855→695 lines) into `ActivityFilters`, `ActivityList`, `ActivityDetailPanel`
- Deleted dead `activity-minimums-config.tsx` (447 lines, zero importers)

**How to Test:**
1. Open `/leagues/{id}/activities` as HOST — page renders identically, search/filter work
2. Expand Configure on an enabled activity — all 4 sections (Basic, Age/Gender Override, Proof Fields, Outcomes) behave as before
3. Save → values persist; Discard → reverts; Toggle on/off works
4. As GOVERNOR — read-only; as PLAYER — non-admin view; non-host API → 403
5. Force a bad league ID → structured `{"level":"error",...}` in server log
Screenshorts's:
Screenshot 2026-07-21 125618
Screenshot 2026-07-21 012505
Screenshot 2026-07-21 012521
Screenshot 2026-07-21 012617

**Checklist:**
- [x] Tested locally and it works
- [x] `pnpm build` passes
- [x] Branch is up to date with `develop`
- [x] No files modified outside scope of this task


blackhathckr (changes requested):
Solid cleanup, verified: all 37 `as any` casts are gone from the activities route (0 remain — the leftover lint hits are pre-existing `: any` annotations, not casts), the duplicated `.eq('league_id', ...)` in activity-minimums is a good catch, decomposition keeps all state/handlers in the parent (children are presentational), no formatting-only files inflate the diff, and lint across the touched files drops from 67 warnings (develop baseline) to 25. The remaining `console.error`s in `activity-minimums/route.ts` are on untouched lines — out of scope, fine.

Two items before approval:

**1. Two new TS errors in the new `activity-list.tsx`** (invisible to `pnpm build` only because `ignoreBuildErrors` is on):

- `(79,48)`: `activityFromMap.max_value` — the map's value type has no `max_value`. Add `max_value: number | null` to the row type backing `enabledActivityMap` (`LeagueActivityRow` declares everything else but is missing it).
- `(74,13)`: the spread object's `frequency_type` is `FrequencyType | null | undefined` but `Activity` only allows `undefined` — coerce with `frequency_type: activityFromMap.frequency_type ?? undefined` (or widen `Activity`).

A PR whose purpose is de-casting shouldn't add new type errors; same standard we've applied on #680/#679. (Related: new `any` at `activity-list.tsx:22` — type it while you're in there.)

**2. Nit: "Closes #556" will close the Phase 2 umbrella epic on merge.** #556 spans multiple modules (your #679 and #689 also reference it). Use "Part of #556" like #679 does — otherwise merging this one module closes the whole tracking issue.

blackhathckr (changes requested):
Re-reviewed `51d005b` — partial credit: `max_value` on `LeagueActivity` is correct, the "Part of #556" body fix is right, and the intent of de-`any`-ing `age_group_overrides` is good. But the commit trades the 2 old TS errors for **7 new ones** (verified with `tsc --noEmit` on the branch vs develop baseline). Two items:

**1. `activity-list.tsx(74,13)` — the `frequency_type` error is still there, different path.**

The `?? undefined` coercion only lives in the `activityFromMap` branch of the conditional spread:

```ts
...activity,                    // LeagueActivity → frequency_type?: ... | null
...(activityFromMap ? { frequency_type: activityFromMap.frequency_type ?? undefined, ... } : {}),
```

When `activityFromMap` is falsy, the base `...activity` spread still contributes `frequency_type?: 'daily' | 'weekly' | 'monthly' | null`, so the union keeps `null` and the assignment to `Activity` still fails. Hoist the coercion out of the conditional so it applies to both paths, e.g. add `frequency_type: (activityFromMap?.frequency_type ?? activity.frequency_type) ?? undefined` after both spreads.

**2. `activity-minimum-dropdown.tsx` — `Record<string, any>` → `Record<string, unknown>` broke `parseAgeOverrides` (6 × TS18046).**

Lines 206–212 access `value.ageRange`, `value.minValue`, and do arithmetic on an `unknown` with no narrowing. `unknown` isn't the goal of a de-`any` pass — a precise type is. `buildAgeGroupOverrides()` (line ~220) shows the actual shape, so define it once and use it in both directions:

```ts
interface AgeGroupOverrideEntry {
  ageRange: { min: number; max: number };
  minValue: number;
}
```

then `parseAgeOverrides(overrides: Record<string, AgeGroupOverrideEntry>)` (and thread it through `LeagueActivity.age_group_overrides` / the `onMinimumChange` callback type in `activity-list.tsx`).

**Not yours / not blocking:** `activity-toggle-card.tsx(187,15)` also errors, but it's pre-existing on develop (line 195 there) on an untouched line — leave it.

Same bar as before: a typing PR shouldn't net-add type errors. After the two fixes above, targeted `tsc` on the four touched files should show only the pre-existing toggle-card error.

blackhathckr (approved):
Third-round review — both items from round 2 are properly resolved this time, verified against current develop (merged locally, compiled, linted):

- **frequency_type**: coercion hoisted out of the conditional spread so both the base-activity and map-override paths normalise null → undefined. The TS2322 at activity-list.tsx:74 is gone.
- **age_group_overrides**: precise `AgeGroupOverrideEntry` in the shared `league-activity-types` module, threaded through parseAgeOverrides/buildAgeGroupOverrides, the route, hook, toggle-card and activity-list — all 6 TS18046s gone. This is the right shape for a de-any pass.
- Bonus: the develop merge correctly carries min_frequency + submission_cutoff_time into the decomposed components.
- Verified: targeted tsc shows only the pre-existing activity-toggle-card error (untouched line, agreed out of scope); eslint identical to develop baseline (5 pre-existing warnings, 0 new); lib test suite passes (the one failure was a stale develop test unrelated to this PR, fixed separately in #742).

Approving and merging. Good persistence on this one.

PR #686, opened 2026-07-21
feat: display leaderboard last-updated timestamp and switch to SWR-style data fetching

dhjr:
## Summary
This PR implements a Stale-While-Revalidate (SWR) data fetching strategy for the leaderboard to ensure fresh data is shown on page load, and adds a "Last updated" timestamp footnote to the UI.

## Related Issue

Closes #633 

## Type of Change
- [x] Bug fix
- [x] New feature
- [x] UI/UX improvement
- [ ] Documentation

## Changes Made
- **Implemented SWR Data Fetching**: Removed the early return on client cache hits and the `hasInitialData` API bypass in `useLeagueLeaderboard.ts`. This ensures that while cached data provides an instant UI response on navigation, a background fetch is always triggered to retrieve and display the latest scores.
- **Added `lastUpdated` State tracking**: The `useLeagueLeaderboard` hook now tracks the timestamp of the last successful fetch and persists it inside the client cache payload for seamless reloading.
- **Updated UI**: Modified `leaderboard-client-container.tsx` to conditionally display a small, non-distracting "Last updated: <time>" footnote beneath the leaderboard stats to give users confidence in data freshness.

## How to Test
1. Navigate to the Leaderboard and verify that the data loads immediately (if cached) while a background network request is made to fetch the latest scores.
2. Scroll to the bottom of the Leaderboard view and verify the "Last updated: <time>" footnote is visible and correctly formatted.
3. Submit a new activity on the Dashboard, then navigate back to the Leaderboard. Verify that the new points are reflected automatically without requiring a manual click on the refresh button.
4. Ensure the manual refresh button still functions properly and updates the "Last updated" timestamp.

## Screenshots / Recordings

Screenshot from 2026-07-22 00-28-09

## Checklist


blackhathckr (approved):
Verified the SWR flow end to end. Cache-hit path now correctly falls through to a background fetch (the removed early-return was the staleness culprit), `lastFetched` round-trips through the client cache so the footnote survives back/forward navigation, and `isLoading` only trips when there's no data or on a forced refetch — so SSR/cached renders won't flash a skeleton during silent revalidation. The deliberate tradeoff (every mount now hits the leaderboard API even with fresh SSR data) is exactly what #633 asks for. A background-refresh failure only logs — the container doesn't consume the hook's `error`, so cached data stays visible rather than flipping to an error state. Lint on both files: the one exhaustive-deps warning pre-exists on develop (this branch actually removes one missing dep). Approving.

PR #697, opened 2026-07-22
fix: stop console errors when non-members view league dashboard

GitGuru-sudo:
## Summary
  Stop noisy console errors on the league dashboard when the viewer isn't a league member (e.g. a host or admin
  browsing). These viewers legitimately get 404s from the phase and dashboard-summary endpoints, which the client was
  treating as hard errors.

  ## Related Issue
  Part of  #556 

  ## Type of Change
  - [x] Bug fix
  - [ ] New feature
  - [ ] UI/UX improvement
  - [ ] Documentation

  ## Changes Made
  - `use-league-phase.ts`: treat 401/403/404 from `/api/leagues/[id]/phase` as an expected empty state — set `phaseInfo:
  null` and degrade quietly instead of logging a console error.
  - `stats-grid.tsx`: treat 401/403/404 from `/api/leagues/[id]/dashboard-summary` as an expected empty state and return
  quietly.
  - Genuine failures (5xx / network) still throw and log as before.

  ## How to Test
  1. Log in as a platform admin (a user who is not a member of the target league).
  2. Open a league dashboard and check the browser console.
  3. Confirm the "Failed to fetch phase" and "Failed to fetch dashboard summary" console errors no longer appear.
  4. Log in as a normal league member and confirm the summary and phase indicator still render correctly.

  ## Screenshots / Recordings
  
Screenshot 2026-07-22 182613
Screenshot 2026-07-22 182625
Screenshot 2026-07-22 182738
Screenshot 2026-07-22 182920

  ## Checklist


blackhathckr (changes requested):
The mechanism is sound and I verified the important safety property: both early returns still hit their `finally` blocks (`setLoading(false)` in `use-league-phase`, `setIsRefreshing(false)` in `stats-grid`), so nothing gets stuck in a loading flag. `phaseInfo: null` also degrades correctly since the hook already falls back to `phase: 'mobilisation'`. Typecheck clean, no new lint warnings.

Two issues before this can go in:

**1. Blocker — `stats-grid` now fails silently into a permanent skeleton.**

On 404 the early `return` leaves `data` as `null`, and a few lines down:

```ts
if (!data) {
  return <StatsSkeleton ... />;
}
```

So a non-member viewer sees a **loading skeleton that never resolves** — the dashboard looks permanently stuck. That was already true before this PR (the throw also left `data` null), but the console error was the only signal it was happening; removing it makes a broken UI silent. Since the whole point here is "this state is expected", it should render an actual terminal state, not a loader: add an `unavailable`/`notMember` flag set in that branch and render a small empty state ("Summary isn't available for this league" or similar) instead of `StatsSkeleton`.

I suspect this is what your screenshot comment is about — if the cut-off UI you're flagging is this skeleton, that's the fix.

**2. Don't swallow 401 — it isn't "expected".**

In both routes 401 means *unauthenticated* (`getAuthUser` returned null), not "not a league member":

```ts
const authUser = await getAuthUser(request);
if (!authUser) return NextResponse.json({ error: 'Unauthorized' }, { status: 401 });
```

`403`/`404` are the legitimate "viewer isn't a member" codes. Treating 401 as a quiet empty state means an expired session silently renders an empty dashboard instead of surfacing anything — exactly the kind of bug that's painful to diagnose later. Please drop `401` from both quiet lists (keep `403`/`404`), and let 401 fall through to the existing error path.

Nit: "Part of #556" is right, but this is a bug fix rather than Phase-2 cleanup — worth its own issue so it's tracked independently.

blackhathckr (changes requested):
Commit `fe77833` changes what this PR is, and I need it split before it can go further.

**Scope: this PR is no longer reviewable as described.**

It opened as a focused 2-file fix ("stop console errors when non-members view league dashboard"). It is now **43 files, +3,786 / −3,365** — a full admin-panel restructure: new `lib/services/admin/*` modules, 11 new `types/admin/*` files, 9 new `user-analytics` components, 4 `preset-challenges` components, a `data-table` decomposition, and three rewritten admin pages.

The title, summary, and "Changes Made" still describe only the console fix, and the checklist still claims *"I have not modified any files outside the scope of this task"* — which no longer holds. Bundling an unrelated 3.7k-line refactor with a bug fix means neither gets reviewed properly, and it blocks a 2-file fix behind a large-surface change.

**Please split into two PRs:**
1. **This one** — keep only `use-league-phase.ts` + `stats-grid.tsx`, with the two items below fixed. It's small and can merge quickly.
2. **A new PR** for the admin-panel refactor, against #556, with its own description and test notes.

**The two items from my previous review are still unaddressed** (I re-checked on `fe77833`):
- `stats-grid.tsx` — the early `return` on 404 still leaves `data` null, and `if (!data) return <StatsSkeleton />` at line 93 still renders a **loading skeleton that never resolves** for non-members. Needs a real empty/unavailable state, not a loader.
- Both files still swallow **401**. 401 is *unauthenticated* (`getAuthUser` returned null), not "not a member" — 403/404 are the right codes to quiet. As written, an expired session silently renders an empty dashboard.

**On the refactor itself** — I did review it so this isn't a wasted round. It's in good shape mechanically: zero TypeScript errors across all 43 files, no `as any` added, no `console.*` added, no formatting-only churn, lint delta neutral, and the `types/admin.ts` and `services/admin/index.ts` barrels keep existing import paths working so nothing downstream breaks. Once it's in its own PR I'll review it properly on behavioral grounds.

Separately: I've opened **#699** for the admin-panel horizontal-overflow problem your screenshots surfaced (Activities, Tiers, AI Usage — clipped action buttons and table columns at mobile widths, plus the floating avatar overlapping card content). Assigned to you, and answered there rather than here.

blackhathckr (approved):
Re-reviewed `830f219` — both items from my previous reviews are resolved, approving.

**Verified:**
- **401 handling**: dropped from the quiet list in both files; unauthenticated responses now fall through to the error path, so an expired session surfaces instead of silently rendering an empty dashboard. Exactly what was asked.
- **Terminal state**: `unavailable` flag + `StatsUnavailable` card replaces the never-resolving `StatsSkeleton` for non-members on 403/404. The `effectActive` guard on `setUnavailable` also correctly avoids setting state after unmount.
- **Scope**: back to a clean 2-file diff (+47/−2). Thanks for the force-push cleanup.
- **Checks**: applied the diff onto current develop — TS error count unchanged (110 baseline, none in touched files); eslint on both files shows only the 4 pre-existing warnings, nothing new.

One follow-up (not blocking): I can see `refactor/admin-panel-god-file-split` pushed to origin — when you open that PR per our agreed split, reference #556 with "Part of", not "Closes".

blackhathckr (changes requested):
Superseding my earlier approval — one condition from the agreed split is still unmet, and it was half of the deal in my scope review: the console fix stays here, **and the admin-panel refactor gets its own PR**.

The code in this PR remains good (both items verified fixed, no new TS/lint issues) — nothing further to change in the diff itself. What's outstanding:

- The `refactor/admin-panel-god-file-split` branch is pushed to origin but has **no PR opened**. Please open it now, against `develop`, with its own description and test notes as agreed in the split plan, referencing #556 with "Part of" (not "Closes").

Once that PR is up, ping me here and I'll re-approve this one immediately.

blackhathckr (approved):
Re-approving — the one outstanding condition is fulfilled: the admin-panel split landed as #733 (reviewed and merged, references #556 correctly). This PR's 2-file fix itself was already verified in my earlier approval (401 handling + StatsUnavailable terminal state, no new TS/lint issues) and still merges clean with current develop. Merging.

PR #714, opened 2026-07-23
refactor: migrate role check to service utility and add unit tests for reupload submission route

dhjr:
## Summary
Fixes `POST /api/submissions/[id]/reupload` returning a misleading `404 Submission not found` error for valid submission IDs due to an invalid column query on `leaguemembers`.

## Related Issue
Closes #712 

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] UI/UX improvement
- [ ] Documentation

## Changes Made
- Replaced non-existent `role` column with `role_id` in the `effortentry` select query join on `leaguemembers`.
- Separated database query errors (`500 Internal Server Error`) from missing record conditions (`404 Not Found`), preventing PostgREST schema failures from being masked as missing records.
- Updated privileged role evaluation to use `userHasAnyRole` helper service (`['host', 'governor']`) instead of querying missing object properties.
- Added unit tests in `tests/unit/reupload-submission-route.test.ts` covering unauthenticated (401), DB failure (500), missing record (404), and successful reupload (200) scenarios.

## How to Test
1. Run unit tests: `pnpm run test -- tests/unit/reupload-submission-route.test.ts`
2. Send a `POST` request to `/api/submissions/<REJECTED_SUBMISSION_ID>/reupload` with a valid session cookie and payload (`proof_url`, `notes`, `tzOffsetMinutes`).
3. Verify the handler returns `200 OK` with `{ "success": true, "data": ... }`.

## Checklist
- [ ] `pnpm build` passes with no errors. ( Error present in Develop branch. Changes out of scope of this PR. Needs to be seperately addressed.)


blackhathckr (approved):
Verified end to end — this correctly fixes #712, and goes a step beyond the minimal patch. Approving.

**The fix:**
- `role` → `role_id` in the `leaguemembers` select — resolves the `42703` error that was surfacing as a misleading `404 Submission not found`. Verified no stale `leagueMember.role` references remain in the route.
- Correctly recognised that `role_id` is a UUID, not a role name, so a `=== 'host'` comparison won't work — routed the privilege check through the existing `userHasAnyRole()` service instead. Good call.
- **Also fixed my second point:** the error branch is now split — a query error (`fetchError`) returns `500 Failed to fetch submission`, and only a genuinely missing row returns `404`. That masking is what hid this bug for a week; splitting it means the next schema error surfaces honestly.

**Tests:** the new `reupload-submission-route.test.ts` is solid — 4 cases (401, 500-on-db-error, 404-not-found, 200 happy path) and it explicitly asserts the select uses `role_id` and not `role`. Ran via `npm test` (Jest): all 4 pass. Nice regression guard for exactly this class of bug.

**Checks:** `tsc --noEmit` clean on both touched files; merges into develop with zero conflicts.

One out-of-scope observation (do NOT fix here): `userHasAnyRole` in `src/lib/services/roles.ts` has a leftover `console.log('[userHasAnyRole] Checking host ...')`. It predates this PR (came in with the testing-codebase update), so it's not yours to clean up here — flagging only so it can be swept in a separate logging pass. LGTM.

PR #722, opened 2026-07-24
hotfix: hide non-actionable 'Pending verification' badge from host team view

blackhathckr:
## Problem (production)
The host/captain "Team Members" dialog shows a **"Pending verification"** badge on several members, with **no action available** to the host — reported by a host on prod (FFL 2).

## Diagnosis (verified against prod)
The badge renders when `member.is_verified === false`, computed server-side as `email_verified OR phone_verified OR admin_verified`. It is **account verification status only** — it has nothing to do with activities or challenges.

Confirmed against prod (FFL 2 / Fit Warriors): the flagged members genuinely have all three verification flags false (pre-registered accounts that never verified email/phone), but they are **fully participating** — e.g. 18/18/17/8 effort entries logged. So the badge:
- is not host-actionable (a host cannot verify someone else's identity — only the member or an admin can),
- has no functional impact on the member's participation or scoring,
- reads as an alarming amber warning the host can neither resolve nor needs to care about.

## Fix
Remove the badge from `view-team-members-dialog.tsx` (the host/governor team-management view rendered by `TeamsTable`). Verification status remains available to the member and to admins; it is simply no longer surfaced as a non-actionable warning in the host roster.

Pure removal — 8 lines, one file. `Badge` import stays (still used for role chips). `tsc --noEmit` clean.

## Release
Hotfix branched off **main** (file identical across main/staging/qa/develop). After merge, back-merge down: main -> staging -> qa -> develop.


adminmfl (approved):
Approved — production hotfix, isolated removal of a non-actionable status badge. Verified no functional impact.

PR #727, opened 2026-07-24
feat: support weight_loss challenge type in leaderboard calculations and add unit tests

dhjr:
## Summary
Fixes an issue where logging a weight loss result (or other non-standard individual challenge results) updated the team's total points on the leaderboard but left the individual player's challenge points unchanged. This ensures player points and team totals remain consistent for all challenge types, including in ended leagues.

## Related Issue
Closes #709 

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] UI/UX improvement
- [ ] Documentation

## Changes Made
- Updated `calculateLeaderboard` in `src/lib/services/leaderboard-logic.ts` to include `weight_loss` and `tournament` challenge types when computing `memberChallengePoints`, and decoupled individual challenge point attribution from team membership checks so players without teams also receive individual points.
- Updated the challenge leaderboard API route in `src/app/api/leagues/[id]/challenges/[challengeId]/leaderboard/route.ts` to process `weight_loss` challenges alongside `individual` challenges when building challenge-specific standings.
- Added unit tests in `tests/unit/leaderboard-weight-loss.test.ts` to verify that weight loss challenge submissions credit both individual player totals and team totals consistently.

## How to Test
1. Run the new unit test suite:
   ```
   pnpm run test -- tests/unit/leaderboard-weight-loss.test.ts
   ```
2. Verify that `calculateLeaderboard` returns matching `total_points` for both individual members and team totals when a `weight_loss` challenge submission is present.

## Checklist


blackhathckr (changes requested):
Reviewed the diff, traced the full points pipeline (`calculateLeaderboard` + `syncSpecialChallengeScores` + `upsertTeamScores`), and ran the new test locally — it passes. The member-attribution fix itself is right: decoupling `memberChallengePoints` from the `teamId && validTeamIds` gate correctly credits team-less players, and adding `weight_loss` to the challenge-leaderboard route's individual branch is correct. Two items before this goes in:

**1. Blocker — team points for weight_loss are double-counted in production, and the test masks it.**

Trace: when a member's final weight is scored, `scoreWeightLossMember` → `syncSpecialChallengeScores` → `upsertTeamScores` writes the team's summed points into `specialchallengeteamscore` (it does this for *all* challenge types, special-challenge-score.ts:283). Meanwhile in `calculateLeaderboard`, the same approved submission also lands in the non-team else-branch and accumulates into `teamChallengePoints`. Then at leaderboard-logic.ts:578:

```ts
const totalBonus =
  (specialChallengeBonus.get(ts.team_id) || 0) +   // synced weight-loss points
  (teamChallengePoints.get(ts.team_id) || 0);      // same submissions again
```

So a member who earned 25 pts shows 25 individually but contributes **50** to the team's `challenge_bonus` — the exact player/team inconsistency #709 is about, just inverted. The unit test doesn't catch it because it mocks `specialchallengeteamscore` as `[]`, which is not the state production will be in after a single scoring run.

Please: (a) verify against dev Supabase (a scored weight-loss league should show the doubled team bonus), (b) dedupe — either exclude challenge types that sync to `specialchallengeteamscore` from `teamChallengePoints`, or filter `specialChallengeBonus` to challenges not already counted from submissions, and (c) add a second test case with the synced team-score row present, asserting the team total equals the member total. I'm aware the same trace applies to plain `individual` challenges approved via the validate route — if you'd rather scope that out, open a follow-up issue, but the weight_loss path is squarely in this PR's remit.

**2. Nit — `isIndividualType` is self-contradicting:**

```ts
const isIndividualType =
  chal.challenge_type === 'individual' ||
  chal.challenge_type === 'weight_loss' ||
  chal.challenge_type === 'tournament' ||
  (chal.challenge_type !== 'team' && chal.challenge_type !== 'sub_team');
```

The final clause subsumes the three explicit ones — they're dead code. Decide which semantics you want: an explicit allowlist (safer — a future challenge type won't silently start crediting members) or the negative catch-all, and keep only that one. Given the tournament inclusion is deliberate here, I'd go allowlist.

**Note, not blocking:** `calculateLeaderboard` counts `['approved', 'pending']` submissions while the sync counts only `approved` — one more consistency wrinkle in the same pipeline, pre-existing; fine to leave, just flagging since this PR is about score consistency.

blackhathckr (approved):
Since this was blocking the weight-loss release train, I implemented the requested changes myself on the branch (rebased onto develop): submission-counted parent challenge ids are now tracked and their synced specialchallengeteamscore rows skipped when building specialChallengeBonus — fixing the team double-count while preserving score-only challenges; isIndividualType collapsed to its effective condition; tests extended to cover the synced-row production state and a score-only challenge (3/3 pass, tsc at file baseline). @dhjr please review the added commit post-merge and flag anything you'd have done differently.

PR #728, opened 2026-07-24
fix(weight-loss config): drop redundant Total Points, open-ended top …

GitGuru-sudo:
## Summary
Removes the redundant Total Points input for weight-loss challenges, renders the top scoring tier as open-ended ("X% and above"), and derives the challenge duration from the selected start/end dates instead of a separate manual field.

## Related Issue
Closes #700

## Type of Change
- [x] Bug fix
- [ ] New feature
- [x] UI/UX improvement
- [ ] Documentation

## Changes Made
- **Total Points hidden for weight loss** — Wrapped the Total Points input in `configure-challenges/page.tsx` with a `challengeType !== 'weight_loss'` condition in both the create and edit dialogs. The server already ignores the submitted value for weight loss (derives points from tiers), so the field was dead input that confused hosts.
- **Open-ended top tier** — Updated `tier-config-section.tsx` to render the last tier as `"X% and above"` with no max input box. Non-final tiers retain the `"to"` range input. The data model already stores the last tier as `max_pct: null`; only the UI was wrong.
- **Duration derived from dates** — Removed the manual `Duration (days)` input from `TierConfigSection`. Added a read-only duration display below the date pickers that computes `differenceInCalendarDays(end, start) + 1`. The derived value is stamped into the config payload on create and edit.
- **Server-side derivation** — Updated both `route.ts` (POST) and `[challengeId]/route.ts` (PATCH) to derive `duration_days` from the effective start/end dates, falling back to the submitted config value, then 30.

## How to Test
1. Open League → Configure Challenges as a host
2. Click Create challenge → choose Type = Weight Loss → verify Total Points field is hidden, tier editor appears
3. Verify the last tier row shows "10% and above = 4 pts" with no empty max box
4. Verify no "Duration (days)" input exists; set Start Date = 2026-08-01, End Date = 2026-08-30 → see "Duration: 30 days"
5. Save the challenge → reopen via Edit → confirm duration matches dates
6. Switch Type to Individual → confirm Total Points field reappears

## Screenshots / Recordings
Screenshot 2026-07-24 155925

Screenshot 2026-07-24 165538
image

## Checklist


blackhathckr (changes requested):
Reviewed against `weight-loss-scoring.ts` and the tier flows — the design is right and the tricky parts are handled well:

- Hiding Total Points for weight_loss matches the server (it derives from tiers and ignores the field) — correct diagnosis of dead input.
- The open-ended top tier UI is **consistent with server validation** (`validateTiers` already rejects a non-null max on the top tier), and I specifically checked the edge cases: `addTier` bounds the previously-open tier before appending, and `removeTier` re-nulls the new last tier's `max_pct` — so there's no way to end up with a hidden, uneditable non-null max that the server would reject. Nice.
- Server-side duration derivation in POST/PATCH (with fallback to existing dates in PATCH) is the right place for it — the client hint text is display-only.

One item before approval:

**The inclusive-days formula is now copy-pasted five times.** `Math.round((end - start) / 86400000) + 1` with a `Math.max(1, …)` guard appears in: `configure-challenges/page.tsx` (`durationFromDates`), inline in `create-challenge-dialog.tsx`, inline in `edit-challenge-dialog.tsx`, `challenges/route.ts`, and `challenges/[challengeId]/route.ts`. The page even defines the helper and then the two dialogs re-inline the same expression instead of importing it. Worse, they're not quite identical — `durationFromDates` returns `null` for non-positive spans while the other four clamp with `Math.max(1, …)`, so an end date before the start date currently yields *different* durations depending on the code path (create-dialog hint says "1 days", page payload falls back to the old config value).

Please extract one `durationFromDates(start, end)` into `src/lib/services/challenges/weight-loss-scoring.ts` (or a shared date util), pick one semantics for the invalid-range case, and use it in all five sites — both client and server import it.

**Merge-order heads-up (not a change request):** this PR and #732 both touch `tier-config-section.tsx` and `challenges/route.ts` — whichever lands second needs a rebase; flag me for re-review after rebasing.

blackhathckr (approved):
Re-reviewed the fix commit — the single item is resolved properly: one durationFromDates in weight-loss-scoring.ts used at all five sites, divergent invalid-range semantics unified on null-with-fallback (with the 400 guard on inverted ranges before persisting — better than the silent Math.max(1) clamp), and the dialog hint copy handles the inverted case honestly. Verified merge with current develop compiles clean and all 55 challenge tests pass. Approving and merging.

PR #729, opened 2026-07-24
fix(weight-loss): accept valid final weight and show it without refresh

GitGuru-sudo:
## Summary
Accept valid final weight on the first attempt and show it immediately without requiring a manual refresh. Drive the participant result state off the persisted end log instead of gating solely on the scored submission.

## Related Issue

Closes #701

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] UI/UX improvement
- [ ] Documentation

## Changes Made

- Render result card off `hasEnd` instead of `hasEnd && result` so a logged final weight is never hidden behind a missing result
- Disable log buttons when input is empty or final weight already logged, removing the empty-field re-click that produced the misleading "Enter a valid weight" toast
- Capture and throw on submission upsert error in `weight-loss-finalize.ts` instead of silently swallowing
- Wrap scoring call in try/catch in `weight-log/route.ts` so a scoring failure returns a soft warning instead of a 500 after the end row is already persisted

## How to Test

1. Log in as a league member in an active weight-loss challenge
2. Enter `60` → Save baseline, then enter `59` → Log final weight
3. Verify result shows immediately with no refresh and no error toast
4. Verify "Log final weight" button is gone/disabled after logging
5. Verify log button is disabled when field is empty

## Screenshots / Recordings

https://github.com/user-attachments/assets/67c907fb-3b42-4ea9-9a40-f96f64c6f907

## Checklist


blackhathckr (changes requested):
Good direction — driving the result card off `hasEnd` (with `result?` fallbacks to the raw logs) is the right fix for the refresh bug, the disabled-button + `weightValid` guard kills the misleading toast, and surfacing the upsert error in `scoreWeightLossMember` instead of swallowing it is correct. Verified the flow against `weight-log/route.ts`. One real gap before this can go in:

**1. Blocker — the scoring-failure path strands the participant permanently, and the warning text promises a retry that doesn't exist.**

When `scoreWeightLossMember` throws, the POST returns `success: true` with `warning: 'Final weight saved; score will be recalculated shortly.'` But:

- `scoreWeightLossMember` is called from exactly one place — this POST's `logType === 'end'` branch. Nothing else ever scores.
- The end log is already persisted, so any retry POST now 409s at "Final weight already logged" (route.ts:114) and never reaches scoring again.
- Net effect: no submission row is ever created, GET keeps returning `result: null`, and the participant sees "Final weight recorded. Your score is being calculated." **forever**.

Cleanest fix, in scope for this PR: **self-heal in the GET handler** — in the member branch, if the logs contain an `end` entry but the `challenge_submissions` lookup returns null, call `scoreWeightLossMember` (it's an idempotent upsert) before responding. Then the very next load genuinely recalculates and the warning text becomes true. A host/governor variant isn't needed — their table reads all submissions, and each member's next visit heals their own row.

**2. Minor — `warning` is never surfaced client-side.**

`logWeight` reads `json.success`/`json.error` but ignores `json.warning`, so the user gets a plain "Final weight logged" success toast even when scoring failed. Add `if (json.warning) toast.info(json.warning)` (or similar) after the success toast. Low priority once item 1 makes the state self-healing.

**Heads-up on merge order (not a change request):** this PR, #730, and #731 all rewrite `participant-view.tsx`, and #731 also touches `weight-log/route.ts`. Whichever merges first, the other two will need a rebase — flag me for re-review after rebasing rather than resolving conflicts silently.

Answering your video question in a separate comment.

blackhathckr (approved):
Re-reviewed the fix commit — all items resolved: GET self-heals a missing score when an end log exists (idempotent upsert, guarded, re-reads after), `json.warning` now surfaces as a toast, and the synthetic weight-loss submission no longer leaks a View Proof link (guard exactly as discussed, badge/date kept). Screenshot confirms. Merging first of the batch; #730/#731 rebase on top.

PR #730, opened 2026-07-24
feat(weight-loss): show log history, progress graph, and live points …

GitGuru-sudo:
## Summary
Render the player's weight log as a list and a weight-vs-date line graph (baseline blue, latest red) above the current-weight input, both driven by the logs already returned by the API. Add a non-persisting live prediction that scores the typed weight client-side from the returned tiers, plus a scoring reference that expresses every tier — including zero points — as a concrete target weight relative to the baseline. Keeps the existing server prediction line. No API changes.

## Related Issue
Closes  #702 

## Type of Change

- [ ] Bug fix
- [X] New feature
- [ ] UI/UX improvement
- [ ] Documentation

## Changes Made
Render weight logs as date · weight · type list (newest first)
Add LineChart (recharts) above current-weight input — baseline blue, latest red, intermediates grey
Add live points prediction on keystroke via computePctLoss/pointsForPct, no write or network call
Add scoring targets translating each tier's min_pct to a concrete target weight, including "same or heavier → 0 pts"
Retain existing server prediction messag
## How to Test
1. Set baseline, confirm history shows one row and graph one blue dot
2. Type a weight → instant live preview, no network request
3. Type baseline or higher → "0 pts" / "at or above baseline"
4. Clear field → preview gone
5. Log progress → graph adds red latest point, history adds second row
6. Targets box shows reachable weights for every tier + 0 pts row
## Screenshots / Recordings
image
Screenshot 2026-07-25 011045
Screenshot 2026-07-25 011412
Screenshot 2026-07-25 012512
Screenshot 2026-07-25 014333

## Checklist


blackhathckr (changes requested):
Nice feature work — the chart config is sensible (sorted data, `dataMin/dataMax` domain, animation off, custom dots keyed properly), the live preview correctly reuses `computePctLoss`/`pointsForPct` from the shared scoring module instead of reimplementing the math, and its `pct <= 0` branch matches `tierForPct`'s "no points at or above baseline" semantics. Host table unit header is a good touch. One correctness issue and two smaller items:

**1. Blocker — the "Scoring targets" card contradicts actual scoring for a `min_pct: 0` tier (which is the default config).**

`DEFAULT_TIERS` starts with `{ min_pct: 0, max_pct: 5, points: 2 }`. For a 100 kg baseline the card renders:

> Reach **100.0 kg or below** → 2 pts
> Same weight or heavier than 100 kg → 0 pts

Both lines claim 100.0 kg, with different points — and the first is wrong: `tierForPct` returns `null` for `pctLoss <= 0` (weight-loss-scoring.ts:31), so at exactly baseline you score 0. "Or below" is only accurate for tiers with `min_pct > 0` (where `pctLoss >= min_pct` includes the boundary). Fix the `min_pct === 0` case specifically — e.g. render it as "Any loss below {startWeight} {unit}" or "Lose anything → 2 pts" — and keep "or below" for the rest. Your own live preview already gets this boundary right; the targets card should agree with it.

**2. Duplicate parsing of the same input.** `parsed` (line ~76) and `typed` (line ~110) are both `Number(weight)`, and `livePreview`'s guard re-states `weightValid` term by term. Use `parsed`/`weightValid` in the preview and delete `typed`.

**3. Result card dropped the percentage.** The completed view now shows absolute "Weight lost: 1.5 kg" but no longer the `pct_change` — yet the percentage is what actually determines points, and the host table still speaks in %. Show both (e.g. "1.5 kg (1.8%)").

**Merge-order heads-up (not a change request):** this PR re-implements several of #729's changes on the same lines of `participant-view.tsx` (`weightValid`, the `hasEnd` gate, the result-null fallback) with slightly different code. #729 also has a pending change (self-healing the missing score on GET) that affects the same "being calculated" state you render here. Land #729 first, then rebase this on top and drop the duplicated hunks — and flag me to re-review after the rebase.

blackhathckr (approved):
Re-reviewed the fix commit — all three items resolved: the min_pct=0 tier now reads 'Lose any weight below {baseline}' (matching tierForPct's strict-loss semantics), duplicate typed parse removed in favour of parsed/weightValid, and the result card shows kg + % with a sane fallback chain. I rebased onto develop myself (post-#729/#731) and combined the conflicting regions: livePreview + chart/history from this PR, canLogFinal window gating and weightValid disabling from the merged PRs, richer startVal/endVal result card kept. tsc + eslint clean vs develop baseline. Approving and merging.

PR #732, opened 2026-07-25
feat(challenges): weight loss as a ready-to-launch default challenge

GitGuru-sudo:
## Summary
Ship weight loss as a ready-to-launch default challenge in the library with editable tier config, dates, and visual "editable" treatment in the launch wizard.
## Related Issue
Closes  #704 

## Type of Change

- [ ] Bug fix
- [X] New feature
- [ ] UI/UX improvement
- [ ] Documentation

## Changes Made
New migration — seeds weight_loss preset into specialchallenges with defaults (4 pts, 30 days), idempotent
challenge-launch-service.ts — widened ChallengeLaunchTemplate.challengeType to include 'weight_loss'; added 
weightLossConfig; populated in listLaunchTemplates; launchChallengeFromTemplate validates tiers, derives duration/points, persists config on insert
challenge-launch/route.ts — accepts weightLossConfig in POST body and passes to service
challenge-launch-wizard.tsx — renders editable date inputs + TierConfigSection with accent/badge for weight loss; sends weightLossConfig on launch; non-weight-loss flow unchanged
tier-config-section.tsx — optional highlightEditable prop adds accent class to tier range/point inputs
challenges/route.ts GET — includes config in select and mapped payload so edit dialog pre-fills launched tiers
## How to Test
1. Run the migration, log in as host → Configure Challenges → Launch a Challenge
2. Confirm Weight Loss Challenge card appears among defaults in Step 0
3. Step 3 — tier editor shows defaults (0–5→2, 5–10→3, 10+→4); date/tier blocks accented with Editable badge; name/description/rules are plain
4. Edit tiers + date → Launch → success toast
5. Open Edit on launched challenge → tiers and dates match what you set
6. As a player → scores by the launched tiers
7. Non-weight-loss template → still shows single Scoring input + static dates
## Checklist


blackhathckr (changes requested):
Solid feature PR — I verified the pieces that could bite and they hold up:

- **Migration is correct against the real schema**: `recommended_scoring`/`suggested_duration` exist (20260706_challenge_template_completeness), `lifecycle_stage` exists (20260705), and the `challenge_type` CHECK already allows `weight_loss` (20260630) as your comment claims. Idempotent guard is right. Leaving `proof_mechanism` NULL is appropriate — no proof step for weight loss.
- **Server doesn't trust the client**: `launchChallengeFromTemplate` re-validates tiers via `validateTiers`, derives `duration_days` from the dates, and computes `total_points` from `maxTierPoints` instead of accepting a client value. Good.
- Launch wizard flow (draft reset on open/template change, `launchingRef` double-click guard preserved) looks right.

Three items before approval:

**1. New `as any` in `challenges/route.ts`:** `config: (c as any).config ?? null`. You added `config` to the select — extend the row type instead of casting. Same standard we've held on #679/#680/#685: no new `as any`. Related question: I don't see anything in this PR's client code consuming `config` from this GET — if it's groundwork for the edit flow in #728, say so in the description; if nothing reads it, drop the field from this PR.

**2. That's now the 6th copy of the inclusive-duration formula.** `Math.max(1, Math.round((end - start) / 86_400_000) + 1)` — I've asked in #728 to extract a shared `durationFromDates` helper into the scoring module; use it here too instead of adding another inline copy. Coordinate with that PR on where the helper lives.

**3. Hoist `EDITABLE` and `EditableBadge` out of the component body.** `EditableBadge` is a component type re-created on every render of `ChallengeLaunchWizard` — it's stateless so nothing breaks today, but it defeats reconciliation and it's the pattern we've been cleaning out elsewhere. Both are constants; define them at module scope.

**Merge-order heads-up (not a change request):** #728 restructures `tier-config-section.tsx` (hides the top tier's max input, removes the duration field) and touches `challenges/route.ts` — both files overlap with this PR. Whichever lands second must re-apply its changes onto the new layout (your `editCls` must land on #728's restructured inputs), not just take one side of the conflict. Flag me for re-review after the rebase.

blackhathckr (approved):
Re-reviewed after your review-fix commit (typed config read, shared durationFromDates, hoisted EDITABLE/EditableBadge) — all three items addressed. I rebased onto develop (post-#728/#729/#730/#731) and finished the integration: re-applied editCls onto #728's restructured tier layout (open-ended top tier keeps the highlight on the visible inputs), deduplicated durationFromDates in favour of #728's null-returning version with an explicit inverted-range guard in the launch service, and added the new required weightLossConfig field to the test factory. tsc clean, all challenge tests pass. Approving and merging.

PR #734, opened 2026-07-27
Hotfix: blank host report, analytics naming removed players, report/rest-day UX

blackhathckr:
## Summary
Production hotfix for the issues reported from FFL2 (all root-caused against the prod database and main branch):

1. **Blank "League Activity Summary" PDF** — `leaguemembers → users` join in `host-report-collector.ts` was ambiguous on prod (three FKs to `users`: `user_id`, `created_by`, `modified_by`), PostgREST returned PGRST201, the error was silently swallowed, and the report rendered with zeros despite 765 effort entries in the window. Fixed with `users!leaguemembers_user_id_fkey!inner` + the error is now surfaced. Also coalesces the effortentry PK (`id` on prod vs `effortentry_id` elsewhere).
2. **Analytics naming removed players** — removal only nulls `team_id`; analytics counted all `leaguemembers`. Top/bottom performers and the "inactive for 5+ days" alert now exclude team-less members (same rule the leaderboard uses), with a guard so pre-team leagues still show full lists.
3. **Host report placement** — "Download Report" now lives at the bottom of the league dashboard while the league is running and moves to the header only after the league ends.
4. **My League Summary** — "View Report" (opens the PDF inline) is now the primary action; Download is secondary.
5. **Scheduled-challenge clarity** — Configure Challenges cards now state the challenge goes live automatically on its start date, so hosts stop looking for a "launch" button after editing.
6. **Rest days relabel** — backend has always treated `rest_days` as a league total (`rest-days/route.ts:51`); quick-start and admin settings said "per Week". Relabelled to "total for the league", cap raised 7 → 60 (FFL2 itself stores 8).
7. **Scoring formula copy** — plain-language descriptions matching what `rr-calculator.ts` actually does.

## Test plan
- [x] `tsc --noEmit`: 142 errors, identical to main baseline (0 new)
- [x] eslint on all 8 touched files: 0 errors, all warnings pre-existing
- [x] Disambiguated members join verified directly against prod REST (returns all 40 FFL2 members; bare `users!inner` reproduces PGRST201)
- [ ] Manual: host dashboard report placement before/after league end
- [ ] Manual: View Report opens inline PDF; Download still works


adminmfl (approved):
Root causes verified against prod DB; TS/lint clean vs baseline. Approving for hotfix merge.

PR #761, opened 2026-07-28
Hotfix: leaderboard counts only approved challenge submissions

blackhathckr:
## Summary
Prod report: a player showed 24 points in 22 days (FFL2). Verified against prod data:
1. Activity points were clean — 22 entries, one per day, no duplicates
2. +2 came from a No Sugar challenge submission made on day 1 of the 7-day challenge (a second member has the identical premature +2) — Submit Proof is available from go-live (tracked: #747/#748)
3. Submissions auto-approve with full points on upload since 53a6bd0b — no review ran; reviewer fields empty on both rows (policy revisit tracked in #748)
4. calculateLeaderboard also counted PENDING submissions, with a fallback to full total_points — so a data-only correction could not remove the points

This PR fixes (4): only approved challenge submissions score. The data correction (both premature submissions flipped back to pending for post-Aug-1 review) is applied separately. Expected after both: player 24 → 22; both affected teams −2.

Confirmed against the challenge owner's rule: the No Sugar challenge ends Aug 1 and submission is only possible from Aug 2.

## Test plan
- [x] leaderboard unit tests 3/3 pass
- [x] tsc at main baseline (no new errors)
- [ ] Post-deploy + data flip: FFL2 top performer shows 22


adminmfl (approved):
Root cause verified against prod data; approved-only counting is correct — points must follow review.

PR #762, opened 2026-07-28
feat(communications): auto-generated 3-stage league mobilization copy

GitGuru-sudo:
## Summary
Adds a Communications settings section that auto-generates WhatsApp-friendly mobilization copy in three stages (pre-launch, during, post-league) from the league's real context. Hosts can copy, share, edit, or regenerate each
stage.
## Related Issue
Closes  #670 

## Type of Change

- [ ] Bug fix
- [X] New feature
- [ ] UI/UX improvement
- [ ] Documentation

##Changes Made 
  - New league_communications table (supabase/migrations/20260728_league_communications.sql) — one row per (league,
  stage), upserted on regenerate; cascade-deletes with the league.
  - New prompts module (src/lib/ai/league-comm-prompts.ts) — Coach MFL system prompt enforcing plain-text
  WhatsApp-friendly output (no markdown, ≤2 emojis/stage), a context builder, a 3-stage parser, and a conservative
  markdown stripper.
  - New service (src/lib/services/ai/league-comm-service.ts) — builds league context from getLeagueById +
  leagueactivities + leagueschallenges, calls the existing aiChat (feature ai-motivate, BYOK 3-tier), and
  lists/generates/edits persisted rows.
  - New host/governor-gated API (src/app/api/leagues/[id]/communications/route.ts) — GET list, POST generate/regenerate,
  PATCH inline edit; mirrors the sibling drafts route gate.
  - New UI section (src/components/leagues/settings/sections/communications-section.tsx) — renders the 3 stages as
  copyable cards, reusing the copy/WhatsApp/native-share UX from whatsapp-reminder-button; lazy auto-generates on first
  open when no rows exist.
  - Registered the section — added the communications group to settings-groups.ts (Essentials grid, megaphone icon) and
  the case in settings-section-page.tsx.
  - Does not touch league creation, the AI Manager draft pipeline, the AI client, or auth/RBAC beyond reusing the
  existing host/governor check. Mobile parity tracked separately in mfl-app#222.
##How to Test
  1. Apply the migration supabase/migrations/20260728_league_communications.sql to your dev DB, run pnpm dev, and log in
  as a host/governor of a league that has a name, dates, ≥1 team, and ideally ≥1 activity + ≥1 challenge.
  2. Go to League → Settings and confirm a Communications card (megaphone icon) appears in the Essentials grid.
  3. Open Communications for a league with no prior comms — confirm a brief loading state, then 3 cards (Pre-Launch /
  During / Post-League) appear populated with text referencing the league name, dates, and team count. Reopen the
  section → content is already there (persisted, no regeneration).
  4. Click Copy on a stage → clipboard holds the exact text (icon flips to a check ~2s). Click WhatsApp → opens wa.me
  with the encoded text. Click Share (native) → system share sheet.
  5. Click Edit on a stage → textarea prefilled; change text → Save → card updates and persists on reopen. Click
  Regenerate → all 3 stages refresh (note: regenerate overwrites edits).
  6. Regression: open the AI Manager Communication tab (drafts/library still work), open AI Coach (chat works), and
  create a new league via the normal flow (creation is not slowed and succeeds). As a non-host member, hit
  /api/leagues/[id]/communications → expect 403.

##Screenshots / Recordings
Screenshot 2026-07-28 221233
  Screenshot 2026-07-28 220947
Screenshot 2026-07-28 221223

  Checklist


blackhathckr (changes requested):
Good feature implementation — the 3-stage comm generation with edit/copy/share is well-structured. The prompt engineering is solid (plain text, WhatsApp-friendly, marker-based parsing). A few items need fixing:

**Must fix:**

1. **`catch (err: any)` in the API route** (route.ts lines ~52, ~82) — use `catch (err: unknown)` and narrow with `err instanceof Error ? err.message : 'Unknown error'`. Same pattern in `communications-section.tsx` line ~68. `any` in catch blocks defeats type safety and is flagged by `@typescript-eslint/no-explicit-any`.

2. **`custom_activities(name)` should be `custom_activities(activity_name)`** — In `league-comm-service.ts` line ~35, the select joins `custom_activities(name)` but the column is `activity_name` based on how every other query in the codebase references it (see `host-report-collector.ts:59`). The `(name)` join will return null and activities won't be included in the prompt context. Verify and fix.

3. **`stripMarkdown` regex is too aggressive** — `league-comm-prompts.ts:82` — `/[*_\`#>-]/gm` strips ALL hyphens from the text (including legitimate ones like "mid-league", "well-being", "pre-launch"). The dash `-` in the character class should either be escaped or this should only strip markdown-specific patterns (lines starting with `- `, `> `, `# `, `**text**`, etc.) rather than individual characters.

4. **Missing `maxDuration` export** — The POST handler calls `aiChat` which can take time. Add `export const maxDuration = 60` to avoid Vercel timeout on cold starts, same as we just fixed for report/analytics routes.

**Nit (should fix):**

5. **Auto-generate on first load** — `load(true)` in the useEffect auto-generates communications when the section is first opened (if empty). This means just *navigating* to the Communications settings tab triggers an AI call and costs tokens. Consider showing a "Generate" button instead and only auto-generating if data already exists. Or at minimum, add a confirmation prompt.

6. **Migration file** — The migration creates the table but doesn't add RLS policies. Verify that the `league_communications` table is accessible via the service role client (which bypasses RLS) and confirm no player-facing queries hit this table directly.

blackhathckr (approved):
Re-review after 7c95b81. All 4 must-fix items resolved:

1. **catch (err: any)** → Fixed: all catches now use `catch (err: unknown)` with `err instanceof Error` narrowing
2. **custom_activities join** → Fixed: query now uses `custom_activities(activity_name)` matching the codebase pattern
3. **stripMarkdown too aggressive** → Fixed: replaced the character-class regex with targeted patterns — only strips markdown syntax at line starts (\`# headers\`, \`> blockquotes\`, \`- bullets\`), bold/italic markers, and backtick code. Hyphens within words like "mid-league" are preserved.
4. **maxDuration** → Fixed: \`export const maxDuration = 60\` added to the route

Also verified: no auto-generation on first load — empty state shows a "Generate" button. RLS enabled with no policies = only service role access, which is correct since the API uses \`getSupabaseServiceRole()\`.

PR #763, opened 2026-07-28
feat(challenges): split Save and Launch in challenge creation

GitGuru-sudo:
## Summary
Make it explicit that creating a challenge does not publish it. The create dialog now offers two actions in the same window: Save Draft (creates a draft, unchanged) and Launch (creates the challenge live/scheduled on the chosen dates). Launch is disabled until name, dates, and points are valid. Rename the draft launch affordances — the card CTA and the date-setting dialog — from "Finish Creation"/"Activate Challenge" to Launch, so a saved-but-unlaunched challenge is launched later from the same place and no "Confirm"/"Finish" wording remains in the flow. No status-model or API changes (POST already accepts an optional status).

## Related Issue

Closes #705

## Type of Change

- [ ] Bug fix
- [X] New feature
- [ ] UI/UX improvement
- [ ] Documentation

## Changes Made
create-challenge-dialog.tsx — footer now shows Save Draft + Launch side-by-side; Launch is gated on name + start + end + points being valid; a hint line explains why Launch is disabled or what each action does
finish-dialog.tsx — renamed title from "Finish Challenge Creation" → "Launch Challenge", description from "Set dates to activate this challenge" → "Set the dates and launch this challenge", primary button from "Activate Challenge" → "Launch"
challenge-card-grid.tsx — draft card CTA renamed from "Finish Creation" → "Launch"
configure-challenges/page.tsx — handleCreateChallenge reads data-intent to distinguish Save vs Launch; Launch passes a derived status (active/scheduled) in the POST payload; added createSubmitting flag to disable both buttons during the request

## How to Test
1. Log in as a host/governor, go to League → Configure Challenges → Create Custom Challenge
2. Save Draft: fill only a name → click Save Draft → confirm the card appears with a Draft badge and a Launch button
3. Launch later: on the draft card click Launch → set dates → confirm the card flips to Scheduled (future) or Active (today)
4. Launch direct: open create dialog, fill name + start + end + points → click Launch → confirm the card lands directly in Scheduled/Active (no draft stop) with a single toast
5. Launch gating: with an empty form, Launch is greyed out; add only a name → Save Draft enables but Launch stays disabled; add dates + points → Launch enables
6. Regression: verify the template wizard's "Launch Challenge" and the Edit dialog's "Save Changes" are unchanged; verify no button reads "Confirm", "Finish Creation", or "Activate Challenge"

## Screenshots / Recordings
Screenshot 2026-07-28 225211
Screenshot 2026-07-28 225349

## Checklist


blackhathckr (changes requested):
Good implementation of the Save/Launch split. The UX is clear — Save Draft for incomplete challenges, Launch for ready ones. A few issues:

**Must fix:**

1. **`dataset.intent` approach is fragile** — In `create-challenge-dialog.tsx`, clicking Save Draft or Launch both set `form.dataset.intent` and then submit the form. But `handleCreateChallenge` reads `(e.currentTarget as HTMLFormElement).dataset.intent` — and `e.currentTarget` for a `type="submit"` button inside a form is the **button**, not the form. The cast to `HTMLFormElement` will give you the button element's dataset (which is empty). You need `e.currentTarget.closest('form')?.dataset.intent` or `(e.target as HTMLElement).closest('form')?.dataset.intent`.

    Actually, looking more carefully: the `onClick` handler sets the dataset on the *form* (`e.currentTarget.form`), then the browser fires the form's `onSubmit`. In `onSubmit`, `e.currentTarget` IS the form. So the flow is: onClick sets dataset → form submit fires → onSubmit reads dataset. This works **but** there's a race condition: if the user clicks rapidly, the second click's onClick could overwrite the dataset before the first onSubmit reads it. Use a ref instead:

    ```tsx
    const intentRef = useRef<'save' | 'launch'>('save');
    ```

    Or simpler: pass the intent as a parameter to separate handlers.

2. **`launchStatus` should handle missing dates** — In `configure-challenges-page.tsx` line ~295, when `intent === 'launch'`, the code computes `launchStatus` from `createForm.startDate`. But if `startDate` is empty string (which passes the earlier validation because the `if` check is `createForm.startDate && createForm.endDate && startDate > endDate`), `new Date('')` is `Invalid Date` and `start.getTime()` returns `NaN`, making the comparison always false → defaults to `'scheduled'` even for today's date. Add a guard:

    ```tsx
    if (intent === 'launch' && (!createForm.startDate || !createForm.endDate)) {
      toast.error('Start and end dates are required to launch');
      return;
    }
    ```

3. **Both `onSubmit` and `onLaunch` are the same handler** — In `configure-challenges-page.tsx` line ~862:
    ```tsx
    onSubmit={handleCreateChallenge}
    onLaunch={handleCreateChallenge}
    ```
    But in `CreateChallengeDialog`, `onLaunch` is called directly via `onPress` on the Launch button — it doesn't go through the form's `onSubmit`. So `e` in `onLaunch` is a **MouseEvent/PointerEvent**, not a FormEvent. The handler does `e.preventDefault()` and `(e.currentTarget as HTMLFormElement).dataset.intent` — both will fail or be no-ops on a PointerEvent. The Launch button should be `type="button"` (not `type="submit"`) and call a separate handler that explicitly sets intent.

**Nit:**

4. **Scheduled vs active distinction is good** — computing `launchStatus` based on whether start date is today or future is the right call. Just make sure it's documented that "scheduled" means "will auto-activate on start date" — does the backend handle that transition?

blackhathckr (approved):
Re-review after c989068. All 3 must-fix items resolved:

1. **dataset.intent race condition** — Fixed: intent is now passed as a function parameter (`handleCreateChallenge('save', e)` / `handleCreateChallenge('launch')`)
2. **Missing date guard for launch** — Fixed: explicit check added (`if (intent === 'launch' && (!createForm.startDate || !createForm.endDate))`)
3. **onSubmit/onLaunch same handler issue** — Fixed: Launch button is `type="button"` with `onClick={onLaunch}`, Save Draft goes through form submit — separate paths, no type mismatch

`createSubmitting` state properly prevents double-submits. `launchStatus` logic (scheduled vs active based on today's date) is clean.

PR #767, opened 2026-07-29
Hotfix: dashboard timeline shows raw UUID for custom-activity entries

blackhathckr:
## Summary
Prod report (host screenshot): activity history rows display a raw UUID instead of the workout name.

Root cause: entries store the custom_activity_id in workout_type (by design — other routes resolve it), but the dashboard summary's leagueactivities query joined only activities(activity_name). Custom activities join custom_activities, which was never selected, so buildActivityPointsMap had no name for any custom activity and buildEntrySubtitle fell back to rendering the raw UUID. 374 entries across prod are affected in display (data itself is fine — no data fix needed).

Fix: select custom_activities(activity_name) and use it in the name/points maps. Join verified live against prod REST (resolves "Daily Standup Update" for the reported UUID).

Note: the "no run rate" part of the report is by design — that league uses the points_only formula, which intentionally shows points without RR. Standard-formula leagues (e.g. FFL2) display RR normally.

## Test plan
- [x] tsc: only the two pre-existing cast errors on these lines (present on main baseline, unchanged)
- [x] dashboard unit tests pass
- [x] Join probed live against prod schema


adminmfl (approved):
Root cause verified; join probed against prod; display-only fix.

PR #771, opened 2026-07-29
P0 hotfixes: weight-loss submission rules, leaderboard filter, custom activity names

blackhathckr:
## Summary
Three P0 fixes requested from prod feedback (Jul 28):

**Weight loss (#743/#744/#745):**
- Final-weight window is now the last day of the challenge or the day after (end..end+1) — the day-before allowance is removed (32 window tests updated, all pass)
- The final weight is overwritable until the submission deadline: the end log updates in place instead of 409ing, and the result card gains an Edit affordance while the window is open
- Instant gratification: the result card congratulates with the earned points and confetti the moment scoring completes (it is synchronous); the "being calculated" state remains only as a failure fallback

**Leaderboard filter (client screenshot):**
- Points-filter selections now apply and close the popover; Apply always closes it (previously it silently did nothing without both custom dates — appeared stuck)

**Custom activity names (#756 systemic):**
- Both entry write paths (player upsert + host manual entry) now resolve and persist custom_activity_name, so no consumer can ever render a raw UUID again. Historical backfill SQL applied separately (374 rows).

## Test plan
- [x] weight-loss scoring tests 31/31
- [x] tsc: no new errors in touched files (baseline preserved)
- [x] eslint: warning counts identical to baseline
- [ ] Post-deploy: log final weight on a test challenge → instant points + confetti; edit until deadline; filter closes on selection


adminmfl (approved):
P0 set verified: window semantics per challenge owner's rule, overwrite-in-place, celebration, filter UX, systemic name fix.

PR #811, opened 2026-07-29
UI/team chat

dhjr:
## Summary

Reworks the team chat header: removes the redundant "Team Messages" heading, promotes the team name above a subtitle that tracks the active filter, renames the filter labels, adds a "Team Messages" filter for member-only messages, and adds an info popover explaining what each filter shows.

## Related Issue

Closes #641 , Closes #642 , Closes #643

## Type of Change

- [ ] Bug fix
- [x] New feature
- [x] UI/UX improvement
- [ ] Documentation

### #641 — redundant heading, filter-aware title

- Dropped the "Team Messages" heading from `messages/page.tsx`; "Team Chat" already lives in the top-left nav.
- The chat header now shows the **team name** as the primary `<h2>` with a secondary line beneath it that reflects the active filter — "Team Chat" on All, "Host Messages" on the host filter, and so on, driven by `FILTER_OPTIONS[].title`.
- Moved the admin view toggle and mobile sidebar trigger into the chat header while it was being restructured.

### #642 — filter labels and the new Team Messages filter

`FILTER_OPTIONS` is now All → Host Messages → Team Messages → For Captains Only → Important.

- "Announcements" → **"Host Messages"**, and it now selects on *sender* (host/governor/creator) rather than `message_type`, which is what the issue asks for.
- "Captains only" → **"For Captains Only"**.
- New **"Team Messages"** filter — messages from team members, i.e. senders outside league leadership, with announcements excluded.
- "All" and "Important" are unchanged.

### #643 — filter info popover

An info icon beside the filter dropdown opens a popover explaining all five categories, listed in the same order as the dropdown so the two can be read against each other.

### Review follow-ups

- **Extracted `getLeagueLeaderIds()`.** The `host_messages` and `team_messages` branches each repeated the same `assignedrolesforleague` select, the same host/governor filter with the same array-or-object `roles` unwrap, the same `leagues.created_by` fetch and the same dedupe — around 28 duplicated lines that had to stay in sync or the two tabs would overlap or strand messages. Both branches now share one function, and its two lookups run in parallel, so the filtered paths also drop a round-trip.
- **Removed the dead `'announcements'` filter.** It previously meant `.eq('message_type', 'announcement')` but had been changed to return every host/governor message under the same name. Nothing passed it — `FILTER_OPTIONS` dropped it and the `MessageFilter` union was the only remaining reference in the codebase — so this is dead-code removal rather than a behaviour change.
- **Added unit tests** for the helper, including the league-creator-as-leader path and the array-shaped `roles` embed. `message-service.ts` previously had no test coverage at all.

## How to Test

Fixture league on dev: `BACKLOG-CHAT-TEST` (`ecc723ab-ced2-4afa-ba3d-5ab1c38e640d`), two teams, 12 messages covering every sender role and visibility. All fixture users log in with password `Test@1234`. The league creator deliberately has **no** `assignedrolesforleague` row, so their messages exercise the creator-as-leader path.

1. As `backlog-chat-playera1@test.local`, open `/leagues/ecc723ab-ced2-4afa-ba3d-5ab1c38e640d/messages`. The header should show **BACKLOG-CHAT-TeamA** on top with **Team Chat** beneath it, and no separate "Team Messages" heading anywhere.
2. Change the filter to **Host Messages**. The subtitle should change to "Host Messages" and the list should show 5 messages — the host's announcement, chat and important message, the governor's message, and the creator's message. The creator's presence is the point: they have no assigned role row.
3. Switch to **Team Messages**. You should see exactly 4 messages, all from team A members, and *no* host announcement, *no* team B message, and *no* captains-only message.
4. Switch to **For Captains Only** as this player — the list should be empty. Log in as `backlog-chat-capa@test.local` and the same filter should show team A's captains-only message.
5. Open the info icon beside the filter dropdown and confirm the five categories are listed in the same order as the dropdown.
6. As `backlog-chat-host@test.local`, toggle **Admin View** on and confirm all 12 messages are visible.

## Screenshots / Recordings

Header and popover changes are visual — worth a look at 375px as well as desktop.

## Checklist


blackhathckr (changes requested):
Good refactor — removing the redundant heading, relocating admin view toggle into the chat header, and adding the info popover are all clean.

**Must fix:**

1. **Filter order mismatch between dropdown and info popover** — The `FILTER_OPTIONS` array orders the filters as: All → Host Messages → **Team Messages** → **For Captains Only** → Important. But the info popover `<ul>` lists them as: All → Host Messages → **For Captains Only** → **Team Messages** → Important. Positions 3 and 4 are swapped. Match the popover order to the dropdown order so users can visually correlate them.

blackhathckr (dismissed):
Re-reviewed after `17fec25`.

**Fixed** — the popover now lists All → Host Messages → Team Messages → For Captains Only → Important, matching `FILTER_OPTIONS` exactly. That was my only blocker from Jul 29.

I also went back over the rest of the diff more carefully this time, since the `team_messages` filter is the functional risk here. Two things I checked and am happy with:

- **Visibility scoping is intact.** `team_messages` falls through to the standard `isHostOrGovernor` / `isCaptain` / else narrowing further down, so a player can't see captains-only or other teams' messages through the new tab. Worth noting because the `captains_only` branch deliberately skips that block, and it would have been easy to land the new filter on the wrong side of it.
- **`.neq('message_type', 'announcement')`** is belt-and-braces rather than load-bearing — `message_type` only holds `announcement` and `chat` on dev, and host/governor senders are already excluded by the preceding `.not(...)`. Harmless, keep it.

Two things still to fix.

### 1. The leader-ID lookup is duplicated verbatim

The `team_messages` branch repeats ~28 lines from the `host_messages` branch with no change: the same `assignedrolesforleague` select, the same `host || governor` filter with the same `Array.isArray(l.roles)` unwrap, the same `leagues.created_by` fetch, the same `[...new Set(...)]` dedupe. Only the final line differs (`.in(...)` vs `.not(..., 'in', ...)`).

That's two copies to keep in sync and two extra round-trips on every request. Pull it out:

```ts
async function getLeagueLeaderIds(supabase, leagueId): Promise<string[]>
```

and have both branches call it, so they can't drift.

### 2. `'announcements'` is now a dead filter with silently changed meaning

The condition became:

```ts
if (filter === 'announcements' || filter === 'host_messages') {
```

Previously `announcements` meant `.eq('message_type', 'announcement')`. Now it returns *every* host/governor message regardless of type — a different result set under the same name.

Nothing actually passes it: `FILTER_OPTIONS` dropped it, and I grepped the codebase — the only remaining reference is the `MessageFilter` union in `message-types.ts:63`. So there's no live bug, but the type still advertises a filter that no longer does what its name says, which is a trap for whoever wires it up next.

Remove `'announcements'` from `MessageFilter` and reduce the condition to `filter === 'host_messages'`.

Fix those two and I'll take another look.

PR #813, opened 2026-07-29
fix: certificate ranking matches leaderboard (#787)

blackhathckr:
## Summary
- Certificate PDF was showing wrong rank (#47) because `ReportRankingsCalculator` was a separate scoring implementation from the leaderboard
- Root causes: dedup by (member_id, date) undercounted entries, `type` field missing from select broke dedup logic, standard formula gave 1 point per entry instead of actual scoring, challenge points handled differently
- **Fix**: Replaced `ReportRankingsCalculator` with `calculateLeaderboard` (the same function the leaderboard UI uses), so certificate rank and points are always identical to what users see on the leaderboard
- Deleted the now-unused `report-rankings-calculator.ts`

Closes #787

## Test plan
- [ ] Generate a certificate for a player in RVKS league — rank should match leaderboard
- [ ] Generate certificate for a player in a league with challenges — points should include challenge points correctly
- [ ] Generate certificate with custom date range — rankings should scope correctly
- [ ] Verify PDF report still renders correctly with all fields


adminmfl (approved):
Verified: replaces divergent ReportRankingsCalculator with the canonical calculateLeaderboard. Type-check clean, no other consumers of the deleted file.

PR #826, opened 2026-07-29
fix(challenges): clarify the Swap action as Replace

GitGuru-sudo:
Summary

  Renames the confusing "Swap" action to Replace and makes all three card actions (Edit / Replace / Delete)
  self-explanatory with tooltips and an always-visible caption, per the decision rule in #706 (Edit + Delete don't cover
  Swap's use case, so clarify instead of remove). Also fixes a latent 500 in the swap route that failed every
  non-weight-loss swap. Verified Edit already supports changing start/end dates via the existing PATCH — no change
  needed there.

  Related Issue

  Closes #706

  Type of Change

  - [x] Bug fix
  - [ ] New feature
  - [x] UI/UX improvement
  - [ ] Documentation

  Changes Made

  - challenge-card-grid.tsx — renamed the card action "Swap" → "Replace"; every action (Edit / Replace / Delete) now has
  a plain-language hover tooltip (previously Swap only had a tooltip when disabled, so an enabled Swap had zero
  context); added an always-visible one-line caption under the action row: "Edit changes dates & details. Replace swaps
  in a different challenge (keeps dates, before submissions). Delete removes it."
  - swap-challenge-dialog.tsx — title "Swap Challenge" → "Replace Challenge"; description now states what it does (swaps
  the challenge for another from the library), that dates stay the same, when to use it vs Edit/Delete, and that it's
  only available before anyone submits; footer button "Swap"/"Swapping..." → "Replace"/"Replacing...".
  - swap/route.ts — fixed a latent 500: the non-weight-loss branch read total_points off the template, but that column
  doesn't exist on specialchallenges (templates carry points in recommended_scoring per #546), so it wrote NULL into the
  NOT NULL leagueschallenges.total_points column and every swap failed. Now reads recommended_scoring with a 0
  fallback, matching the create path. Swap logic, submission guard, and auth gate are otherwise unchanged.
  - Edit-dates acceptance criterion — verified, no code change: PATCH already accepts startDate/endDate, re-derives
  status from the new window, and runs the #550 overlap gate; the Edit dialog already sends both.

  How to Test

  1. Run pnpm dev, log in as a host, open League → Configure Challenges with at least one scheduled challenge (no
  submissions) and one challenge with submissions.
  2. Actions are self-explanatory — on a scheduled card, confirm the middle action reads Replace, the caption line under
  the row explains all three actions, and hovering each button shows a plain-language tooltip.
  3. Replace flow — click Replace → dialog titled "Replace Challenge" explains it swaps the challenge, keeps the dates,
  and when to use it vs Edit/Delete, with "Date slot preserved" showing the current dates. Pick a different template →
  Replace → the card now shows the new challenge with the same dates and its template's recommended points (previously
  this 500'd with "Failed to swap challenge").
  4. Submission guard — on the challenge with submissions, confirm Replace is disabled and its tooltip says it's
  unavailable once submissions exist.
  5. Edit changes dates — open Edit, change start/end dates → Save. Confirm the card shows the new dates, status
  re-derives (e.g. Scheduled → Active if start moved to today), and the overlap warning fires on conflicting dates.
  6. Delete unchanged — delete a challenge, confirm the confirm dialog and removal still work.

  Screenshots / Recordings
Screenshot 2026-07-30 003433
Screenshot 2026-07-29 230343

  Checklist


adminmfl (approved):
Clean PR — correctly addresses #706.

**Swap → Replace rename** is thorough: card action, dialog title, dialog description, footer button, and disabled tooltip all updated. No stale "Swap" references left.

**Bug fix in swap route** is a real catch: `targetTemplate.total_points` doesn't exist on `specialchallenges` (templates carry points in `recommended_scoring` per #546), so every non-weight-loss swap silently 500'd. `Number(targetTemplate.recommended_scoring ?? 0)` correctly matches the create path.

**Tooltips and caption** make the three actions self-explanatory without needing docs. The always-visible caption under the action row is a good touch for host-facing UX.

**Edit-dates verification** — confirmed PATCH already accepts `startDate`/`endDate` and re-derives status. No change needed.

PR #829, opened 2026-07-31
Feat/new host onboarding

dhjr:
## Summary

Adds the "How to run this league" host guide from #671 — a new `/leagues/[id]/guide` route with a 4-step path, progress indicator and skip option — and makes it the default landing page for hosts in place of League Settings.

## Related Issue

Closes #671

## Type of Change

- [ ] Bug fix
- [x] New feature
- [x] UI/UX improvement
- [ ] Documentation

### The guide itself

- **New route** `src/app/(app)/leagues/[id]/guide/page.tsx`, gated to host and governor, with a not-found fallback and an access-restricted state for everyone else.
- **`host-guide-shell.tsx`** — header banner, progress bar, the "don't show again" preference and a "Skip to Settings" escape hatch.
- **Four step components** matching the issue's breakdown: `step-1-setup`, `step-2-invite`, `step-3-launch`, `step-4-wrapup`, plus a shared `step-card`.
- **Step links** go to the real pages — step 1 → settings / configure-challenges / activities, step 2 → invite, step 3 → submissions / manual-entry / rest-day-donations, step 4 → leaderboard / ai-manager / analytics.
- **Sidebar entry** added so the guide stays reachable, and League Settings remains in the sidebar as the issue requires.

### Host landing behaviour

Hosts opening a league now land on the guide instead of League Settings. The affected entry points are the leagues list, the league switcher and the role switcher.

### Progress

Progress is **derived from league state**, not self-reported:

- **Step 2** (players invited) is complete once `member_count > 1`.
- **Step 3** (launched) is complete once `start_date` has arrived.

Steps 1 and 4 have no single reliable signal — setup is a judgement call and wrap-up happens after the league ends — so they stay manually ticked and persist in `localStorage`. Derived and manual completions are unioned, so a host who set the league up on a laptop sees the same progress on their phone.

### Review follow-ups

- **Fixed a type error** in `guide/page.tsx`: `getLeagueById` returns `member_count` as optional while `HostGuideShell` required it.
- **Fixed a hydration mismatch** in `leagues/page.tsx`. The skip flag was read from `localStorage` while building each row's `href`, but that file is prerendered on the server where `localStorage` does not exist — so for a host who had skipped, the server emitted the guide link and the client emitted the settings link, on every row. Now resolved in an effect and held in state.
- **Made the skip flag mean one thing.** It previously sent hosts to `/leagues/{id}` from the leagues list but `/leagues/{id}/settings` from both switchers, so where a host landed depended on which control they clicked. Settings is correct per #671, which frames the guide as replacing League Settings as the default landing.
- **Extracted `src/lib/navigation/host-landing.ts`.** The `localStorage` check was inlined at five call sites across three files, and `league-switcher.tsx` defined the right abstraction but used it at only one of its three sites. Both fixes above now live in one place, with unit tests covering the destination-drift regression and storage being unavailable.

## How to Test

Two fixture leagues on dev, both hosted by `backlog-guide-host@test.local` / `Test@1234`, chosen so derived progress is visible without ticking anything:

| League | State | Expected |
|---|---|---|
| `BACKLOG-GUIDE-A-FRESH` (`e80a3cb0-e3cf-48a3-a6e9-b4d87f80ba02`) | host only, starts in 7 days | **0%**, 0 of 4 |
| `BACKLOG-GUIDE-B-RUNNING` (`25d99eaa-2c91-4bea-9c2d-9c216bfd590c`) | 5 members, started 5 days ago | **50%**, steps 2 and 3 already complete |

1. Log in as the host and open **My Leagues**. Both rows should link to `/leagues/{id}/guide`.
2. Open `BACKLOG-GUIDE-B-RUNNING`. The guide should show **50%** with steps 2 and 3 already ticked, without you having ticked anything.
3. Open `BACKLOG-GUIDE-A-FRESH`. It should show **0%** — no players have joined and the league has not started.
4. Tick **"Don't show this guide by default"** on either league, go back to My Leagues, and confirm that league's row now goes to `/leagues/{id}/settings`. Check the same league via the league switcher and the role switcher — all three must agree.
5. Reload My Leagues with the skip flag set and confirm there is no hydration warning in the browser console.
6. Confirm the other league still opens the guide — the preference is per league.
7. Open the guide on a phone at 375px and confirm progress matches what the laptop showed.

## Screenshots / Recordings

Not attached — the two fixture leagues above are set up so the guide, both progress states and all four steps can be viewed directly on dev.

## Checklist


blackhathckr (dismissed):
Checked this against #671's acceptance list — the feature itself covers it: 4-step flow, progress bar, skip option, League Settings still in the sidebar, and each step links out to the right page (step 1 → settings / configure-challenges / activities, step 3 → submissions / manual-entry / rest-day-donations, step 4 → leaderboard / ai-manager / analytics). The `/guide` page's host+governor gate and not-found fallback are both handled properly.

The problems are around it.

### 1. The PR body describes only half of what's here

The body describes making the guide "fully mobile responsive down to 320px" — that's commit `4cd70cf`. But this PR also contains `24bc926`, which is the **entire feature**: a new `/leagues/[id]/guide` route, `host-guide-shell.tsx`, four step components, `step-card.tsx`, a sidebar entry, and a rewrite of where hosts land in three separate files.

So the largest and riskiest part of a 1,505-line PR — changing the default landing page for every host — is undocumented. The Screenshots section is empty for a brand-new UI feature, and **`pnpm build` passes with no errors is unchecked**.

Please rewrite the body to describe the feature, add screenshots of the guide and all four steps, and confirm the build.

### 2. New type error in `guide/page.tsx:84`

```
src/app/(app)/leagues/[id]/guide/page.tsx(84,9): error TS2322:
  Type 'number | undefined' is not assignable to type 'number'.
```

`HostGuideShell` declares `member_count: number` (required), but `getLeagueById` returns `member_count?: number` (`league-types.ts:135`). The repo has pre-existing `tsc` noise, so this isn't the only error — but it's a new one, from this PR, in a new file, and it's exactly what the unchecked build box would have caught. Either make the prop `member_count?: number` and default it in the shell, or `member_count: league.member_count ?? 0` at the call site.

### 3. Hydration mismatch — localStorage read during render in `leagues/page.tsx`

```tsx
const isSkipped =
  typeof window !== 'undefined' &&
  localStorage.getItem(`mfl_skip_host_guide_${league.league_id}`) === 'true';
const targetHref = isHost && !isSkipped ? `/leagues/${id}/guide` : `/leagues/${id}`;
```

`leagues/page.tsx` is `'use client'`, but Next still prerenders it on the server, where `typeof window === 'undefined'` forces `isSkipped = false`. For a host who has skipped, the server emits `href=".../guide"` and the client renders `href="/leagues/{id}"` — a hydration mismatch on every row.

You already solved this correctly in `host-guide-shell.tsx:52-60`, which reads both keys inside a `useEffect` and sets state. Use that same pattern here.

### 4. The same skip flag sends hosts to two different places

- `leagues/page.tsx` — skipped → `/leagues/${id}` (dashboard)
- `league-switcher.tsx` and `role-switcher.tsx` — skipped → `/leagues/${id}/settings`

One preference, two destinations, so where a host lands depends on which control they used. #671 frames the guide as replacing League Settings as the default landing, so skipping should go to settings. Pick one and apply it everywhere.

### 5. The routing logic is inlined five times

The same `localStorage.getItem('mfl_skip_host_guide_' + id) === 'true'` check appears in `leagues/page.tsx` (×1), `league-switcher.tsx` (×3) and `role-switcher.tsx` (×1).

`league-switcher.tsx:129` even defines the right abstraction:

```ts
const getHostLandingUrl = (leagueId: string) => { ... };
```

and then only uses it at one of that file's three call sites — the other two re-inline the logic. Export one helper (`getHostLandingUrl`) from a shared module and call it from all five places. That also makes fixes 3 and 4 land in one spot instead of five.

### 6. Progress is self-reported and per-device — confirm this is intended

#671 asks for "a progress indicator showing completed vs pending steps". Here, completion is an array the host ticks by hand, stored in `localStorage` under `mfl_host_guide_completed_steps_${id}`, and the skip flag is per-device too. So a host who genuinely completed setup on their laptop opens the guide on their phone and sees 0% with nothing marked done, and the bar reflects self-reporting rather than the league's actual state.

Real signals are already passed into the shell — `member_count`, `league_capacity`, `phase`, `start_date` — so at least step 2 (players invited) and step 3 (league launched) could be derived. Either derive those two, or say explicitly in the PR body that manual per-device marking is the intended v1 so it's a recorded decision rather than an oversight.

Fix 2-5, sort out the body, and answer 6 — then re-request review.

PR #827, opened 2026-07-30
fix(challenges): gate challenge configuration when the league has end…

GitGuru-sudo:
## Summary
  Hosts could go through the entire challenge creation/configuration flow on an ended league and only find out at
  submission time. This PR surfaces the ended state upfront on the Configure Challenges page and blocks creation at both
  the UI and API level, while keeping review/publish of existing challenges available.

  ## Related Issue
  Closes #707

  ## Type of Change
  - [x] Bug fix
  - [ ] New feature
  - [ ] Refactor (no functional change)
  - [ ] UI/UX improvement
  - [ ] Chore (dependencies, config, CI)
  - [ ] Documentation

  ## Changes Made
  - `GET /api/leagues/[id]/challenges` now returns `data.leagueEnded` (on both the success and empty-fallback paths) so
  the page knows the league state from the existing fetch — no extra request.
  - `POST /api/leagues/[id]/challenges` rejects custom challenge creation with 400 "This league has ended…" when the
  league is `completed` or past the `isLeagueEnded` 33h cutoff — same semantics the final submission check
  (`entries/upsert`) enforces.
  - `POST /api/leagues/[id]/challenge-launch` rejects wizard launches on an ended league with 400, even when the dates
  fall inside the old league window.
  - Configure Challenges page shows a prominent "This league has ended" banner (same `Alert` styling as
  `archive-notice.tsx`) immediately after load.
  - The "Launch a Challenge" card, empty-state "Create Challenge" button, and "Create a custom challenge" section are
  hidden when the league has ended.
  - Belt-and-braces `guardEnded()` toast guard on the wizard opener and `handleCreateChallenge`, in case a stale render
  leaves a button visible.
  - Review, publish, view-proof, and delete for existing challenges remain fully available — only *creation* is gated.

  ## How to Test
  1. Pick a league whose `end_date` is more than 2 days in the past (or set one in Supabase) and log in as its Host.
  2. Open `/leagues/{id}/configure-challenges` → the ended banner appears at the top; the Launch card and Create Custom
  section are gone.
  3. From devtools, `POST /api/leagues/{id}/challenges` with a valid body → 400 "This league has ended. Challenges can
  no longer be created or configured."
  4. `POST /api/leagues/{id}/challenge-launch` with dates inside the old league window → 400 "This league has ended.
  Challenges can no longer be launched."
  5. Open Review on an existing challenge of that league → submissions still load; Publish still works.
  6. Switch to an active league → the page renders exactly as before (Launch card, custom create, no banner).

[...]

PR #830, opened 2026-07-31
feat(challenges): Ready-made terminology + three clear creation paths…

GitGuru-sudo:
## Summary

Standardises challenge terminology on **Ready-made** vs **Custom**, and replaces the single "Launch a Challenge" button with three explicitly labelled creation paths on the configure-challenges page.

## Related Issue

Closes #749

## Type of Change

- [ ] Bug fix
- [x] New feature
- [x] UI/UX improvement
- [ ] Documentation

### Terminology

The user-visible "Default" lifecycle label becomes **Ready-made** across the admin preset table, the launch wizard's template cards, the AI manager library heading, and the migration comment. The DB enum value stays `'default'` — only the display label changes.

### Three creation paths

The configure-challenges page now offers three cards rather than one button plus a separate custom section:

1. **Use a Ready-made Challenge** — opens the wizard in `quick` mode, defaults predefined
2. **Customize a Ready-made** — opens the wizard in `full` mode, all steps editable
3. **Create a Custom Challenge** — the from-scratch dialog

The `quick` / `full` split is what differentiates paths 1 and 2. The steps skipped in `quick` mode are safe to skip: `handleStep2Next` is pure validation with no side effects, and `weightLossDraft` is seeded from the template (or `DEFAULT_WEIGHT_LOSS_CONFIG`) in the `selectedTemplate` effect, so a quick-mode weight-loss launch still carries proper tiers.

### Review follow-ups

- **Merged `develop` and re-applied the #827 gating.** #827 (which fixes #707) merged first, so this branch was rebased onto it and the ended-league gate re-applied: the three-card `<section>` is wrapped in `!leagueEnded`, the ended-league `Alert` is preserved, and every path routes through `guardEnded()`. Without this, resolving the conflict in this PR's favour would have silently regressed #707.
- **Restored keyboard accessibility.** The three cards replaced real `<Button>` elements and were left as bare `<Card onClick>` — mouse-only, not focusable, not announced. A shared `cardAction(label, run)` helper now supplies `role="button"`, `tabIndex={0}`, `aria-label`, Enter/Space activation and the ended-league guard, so all three paths behave as buttons and none can be reached around the gate by tabbing.
- **Reverted the `help/page.tsx` churn.** Only one line is a real edit — the "How do I create a challenge?" answer, rewritten for the three paths. **The remaining diff in that file is the pre-commit Prettier hook**: `help/page.tsx` predates Prettier enforcement and is not format-clean on `develop`, so any commit touching it reformats the whole file. Review that file with `git diff -w`, or just read the single `answer:` string under "How do I create a challenge?".

### Out-of-scope changes, now documented

The wizard also carries a start-date fix that is not part of #749. Flagging it explicitly so QA covers it:

- Prefill switches from `new Date().toISOString().split('T')[0]` to `localYmd(new Date())`. The old code prefilled **yesterday** for IST hosts before 05:30.
- `min={todayYmd}` on the start-date input, a past-date rejection in `handleStep1Next`, and `canGoNext()` case 1 now requires `startDate >= todayYmd`.

The status preview deliberately stays on UTC while the picker uses the host's local day. That is not an oversight: `statusForLaunch` runs server-side on a UTC host, so mirroring it is what makes the preview truthful — an IST host picking their local "today" before 05:30 really will get `scheduled`, and the preview now says so. A comment in the wizard records this.

## How to Test

1. On an **active** league, open configure-challenges. Three cards appear. Tab through them — each takes focus, shows a focus ring, and activates with Enter and Space.
2. Card 1 opens the wizard with steps pre-filled and ready to launch; card 2 opens the same wizard with all steps editable; card 3 opens the custom dialog.
3. On an **ended** league, confirm the orange "This league has ended" alert appears and the three cards are gone. Regression check for #707.
4. In the wizard, confirm the start date prefills to **today in your local timezone**, that earlier dates are rejected, and that the status preview matches the status the challenge actually receives after launching.
5. Help → "How do I create a challenge?" now describes all three paths.

## Screenshots / Recordings

The three-card layout replaces a single button — worth a look, including a keyboard-only pass.

## Checklist


blackhathckr (dismissed):
The terminology sweep itself is clean — I grepped for leftover user-facing "default" / "one-click" / "ready to launch" copy and the only remaining `'default'` hits are shadcn Badge/Button variants and the DB enum, which is exactly right. `LIFECYCLE_DISPLAY` keeping the enum value while relabelling the badge is the correct call. The `quick` / `full` split genuinely differentiates paths 1 and 2, and I checked the skipped steps are safe: `handleStep2Next` is pure validation with no side effects, and `weightLossDraft` is seeded from the template (or `DEFAULT_WEIGHT_LOSS_CONFIG`) in the `selectedTemplate` effect, so a quick-mode weight-loss launch still carries proper tiers.

Four things to fix.

### 1. Collides with #827, and the ended-league gate would regress

#827 is approved and waiting to merge. It gates this exact page: it wraps the "Launch a Challenge" card and the "Create a custom challenge" section in `!leagueEnded`, and adds a `guardEnded()` toast on both the wizard opener and `handleCreateChallenge`.

This PR **deletes both of those blocks** and replaces them with three cards that have no gating at all:

```tsx
<Card onClick={() => { setWizardMode('quick'); setWizardOpen(true); }}>
<Card onClick={() => { setWizardMode('full'); setWizardOpen(true); }}>
<Card onClick={() => setCreateOpen(true)}>
```

Whichever lands second conflicts. If this one lands after #827 and the conflict is resolved by taking this side, #707 silently regresses — a host on an ended league gets all three creation paths back, which is precisely the bug #827 fixes.

Both are yours, so please pick the order explicitly: merge #827 first, then rebase this one and re-apply `!leagueEnded` to the three-card `<section>` plus `guardEnded()` on all three `onClick`s. Call out in the PR body which order you've chosen.

### 2. Undocumented out-of-scope change: start-date validation + UTC→local fix

The wizard picks up a set of changes that have nothing to do with #749:

- `localYmd` import; prefill switches from `new Date().toISOString().split('T')[0]` to `localYmd(new Date())`
- `min={todayYmd}` on the start-date input
- a past-date rejection in `handleStep1Next`
- `canGoNext()` case 1 now `!!startDate && startDate >= todayYmd`

None of it appears in "Changes Made", and the checklist says *"I have not modified any files outside the scope of this task"*. The UTC→local prefill fix looks right and is worth having — the old code really did prefill yesterday for IST hosts before 05:30 — but shipping it unannounced inside a terminology PR means nobody tests it.

Either split it into its own PR against its own issue, or document it fully in the body so QA knows to cover it.

While you're there: the code itself flags an inconsistency this introduces. The status preview still reads

```tsx
{/* mirrors server statusForLaunch (UTC) — keep in sync */}
{startDate > new Date().toISOString().split('T')[0] ? 'scheduled' : 'active'}
```

so `startDate` is now local-based while the comparison stays UTC-based. For a host west of UTC late in the day these disagree about what "today" is, and the preview can show the wrong status. Either make both use the same basis or state why the mismatch is intended.

### 3. The three cards are mouse-only — accessibility regression

They replaced two real `<Button>` elements (`Launch Challenge`, `Create Custom Challenge`), which were focusable and keyboard-operable. The replacements are bare `<Card onClick={...}>` — no `role="button"`, no `tabIndex`, no `onKeyDown`. The page's three primary actions can no longer be reached by keyboard or announced correctly by a screen reader.

You did this correctly in #833 on the challenge card title:

```tsx
role="button"
tabIndex={0}
aria-label={...}
onKeyDown={(e) => { if (e.key === 'Enter' || e.key === ' ') { e.preventDefault(); ... } }}
```

Apply the same treatment here, or wrap each card's content in a `<Button variant="ghost">`.

### 4. `help/page.tsx` is 95% Prettier reflow

113 additions / 47 deletions, and exactly **one** substantive change — the "How do I create a challenge?" answer rewritten for the three paths. Everything else is line-wrapping, trailing commas, and JSX re-indentation, plus four now-unused icon imports removed.

This is the same diff-inflation problem I raised on #513 (Jul 13, item 5). Revert the formatting-only churn and keep just the FAQ answer edit, so the reviewable change is one line instead of buried in 160.

Fix these four and re-request review.

PR #831, opened 2026-08-01
feat: include personal challenge points in total points calculation and update dashboard stats grid layout

dhjr:
## Summary

Breaks a player's score into **Activity Points** and **Challenge Points** with the combined **Total Points** still shown, and labels the Team card's figures explicitly as team figures so they can no longer be confused with the player's own.

## Related Issue

Closes #708

## Type of Change

- [x] Bug fix
- [x] New feature
- [x] UI/UX improvement
- [ ] Documentation

## Changes Made

**Points breakdown.** `processLeagueDashboardSummary` now includes personal challenge points in `mySummary.totalPoints`, and My Summary renders three cells — Total Points, Activity Points, Challenge Points. `points` is reduced from approved effort entries and `personalChallengePoints` from unique challenge submissions, which are disjoint sources, so the sum cannot double-count.

**Team labelling.** My Summary and the Team card each rendered a cell labelled exactly `"Total Points"`, highlighted identically and stacked one above the other — the ambiguity #708 was raised to remove. The Team card now reads **"Team Total Points"** and **"Team Run Rate"**, the latter having the same collision with My Summary's `"Run Rate"`.

**Dead fallbacks removed.** `MySummary` declares `points`, `totalPoints` and `challengePoints` as required non-optional numbers, so `?? mySummary.points + (...)` could never fire and `|| 0` only masked a legitimate zero.

## Reconciliation with the leaderboard

This is the real user-visible win. The dashboard previously showed activity-only points under a "Total Points" label while the leaderboard counted challenge points toward an individual's score, so the two surfaces disagreed by exactly the challenge total for any player with challenge points.

Verified on dev (`BACKLOG-708-TEST`), for a player with 40 activity points and one 25-point challenge:

| Surface | Before | After |
|---|---|---|
| Dashboard → My Summary → Total Points | 40 | **65** |
| Leaderboard → individual total | 65 | **65** |

A second player in the same league with 20 activity points and no challenge submissions reads 20 on both surfaces, confirming the no-challenge path is unaffected.

## How to Test

Fixture league `BACKLOG-708-TEST` (`72bd0a74-09a1-49cb-aeb5-ffc4fc6105a6`) on dev. Password `Test@1234`.

1. Log in as `backlog-708-p1@test.local` and open the league dashboard. My Summary should read **Total Points 65**, **Activity Points 40**, **Challenge Points 25**.
2. Open the league leaderboard and find the same player — their individual total should also be **65**.
3. On the dashboard, confirm the Team card's cells now read **"Team Total Points"** and **"Team Run Rate"**, so no two cells on the page share a label.
4. Log in as `backlog-708-p2@test.local`, who has no challenge submissions. Total Points and Activity Points should both read **20** and Challenge Points **0**.

## Screenshots / Recordings

The label change is visual — worth a look at the two stacked cards on the dashboard.

## Checklist


blackhathckr (dismissed):
The points maths is right. I checked the processor: `points` is reduced from `approvedList` (effort entries, L66-74) and `personalChallengePoints` from `uniqueChallengeSubs` (L115) — genuinely disjoint sources, so `points + personalChallengePoints` is a clean sum with no double-count. I also traced every consumer of `mySummary.totalPoints`: `stats-grid.tsx` is the only one, and before this PR it read `mySummary.points`, so redefining `totalPoints` breaks nothing else. The `isChallengesOnly` early-return at L173-176 stays self-consistent.

But this only does half of #708.

### Must fix 1 — the team/player labelling half of the issue is untouched

#708 has two requirements, and the second is quoted directly from the meeting:

> There was also confusion between **team total points** and **the player's own points** being shown together without clear labelling.

with the acceptance criterion:

> - Team totals are clearly labelled as team figures.

After this PR, `stats-grid.tsx` renders **two cards, both with a cell labelled exactly `"Total Points"`**:

- My Summary (L170): `<StatCell label="Total Points" value={...} highlight />`
- Team card (L243): `<StatCell label="Total Points" value={mySummary.teamPoints...} highlight />`

Two identically-labelled, identically-highlighted "Total Points" cells stacked one above the other is precisely the ambiguity the issue was raised to remove — and the My Summary number just changed, so the two are now more likely to be visibly different values, which makes it worse rather than better.

Relabel the team card's cells as team figures — `"Team Total Points"` (or `"Team Points"`), and ideally `"Team Run Rate"` for the adjacent RR cell, which has the same collision with My Summary's `"Run Rate"`. The card header helps, but the issue explicitly asks for the figures themselves to be unambiguous.

### Must fix 2 — the new fallbacks are dead code

```tsx
value={(
  mySummary.totalPoints ??
  mySummary.points + (mySummary.challengePoints || 0)
).toLocaleString()}
...
value={(mySummary.challengePoints || 0).toLocaleString()}
```

In `src/lib/types/dashboard.ts:30-43`, `MySummary` declares both as **required, non-optional** numbers:

```ts
points: number;
totalPoints: number;
challengePoints: number;
```

So `??` can never fire and `|| 0` only masks a legitimate `0`. It reads as uncertainty about whether the processor sets the field — it always does, on both branches. Reduce to `mySummary.totalPoints.toLocaleString()` and `mySummary.challengePoints.toLocaleString()`.

### Must fix 3 — test gaps

`dashboard-points-breakdown.test.ts` covers one path with one config. Two cases worth adding, both cheap:

1. **Zero challenge points** — `leagueChallengeSubs: []` should give `totalPoints === points`. This is the common case for most leagues and the one most likely to regress.
2. **`isChallengesOnly: true`** — the early-return at L173-176 sets `points`, `totalPoints` and `challengePoints` all to `personalChallengePoints`. Nothing currently pins that, and it's the branch where the three fields intentionally collapse.

### One thing to confirm

Because the dashboard previously showed activity-only points under the "Total Points" label while the leaderboard counts challenge points toward an individual's score, the two surfaces disagreed for any player with challenge points. This change should bring them into line. Please confirm on a league with challenge submissions that My Summary → Total Points now matches that player's leaderboard individual total, and say so in the PR description — that reconciliation is the real user-visible win here and it's worth having recorded.

Fix the labelling, drop the dead fallbacks, add the two test cases, then re-request review.

PR #832, opened 2026-08-01
feat: add min and max submissions per day fields to league schema and API responses

dhjr:
## Summary
Adds `max_submissions_per_day` and `min_submissions_per_day` fields to the `GET /api/user/leagues` endpoint select query and response serialization. This ensures client applications (including mobile) receive host-configured daily submission limits rather than falling back to default values.

## Related Issue
Closes #717 

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] UI/UX improvement
- [ ] Documentation

## Changes Made
- Updated `GET /api/user/leagues` route (`src/app/api/user/leagues/route.ts`) select query and serializer to include `min_submissions_per_day` and `max_submissions_per_day` (defaulting to 1).
- Updated `LeagueWithRoles`, `DbLeagueRow`, `LeagueWithRoleFields`, and `League` interfaces to include `min_submissions_per_day` and `max_submissions_per_day`.
- Updated `getLeagueById`, `getLeaguesForUser`, and `getUserLeaguesWithRoles` queries and serializers in `src/lib/services/leagues/league-query-service.ts` to fetch and include both fields.
- Updated `updateLeague` in `src/lib/services/leagues/league-mutation-service.ts` to allow updating `min_submissions_per_day` and `max_submissions_per_day` for active/post-draft leagues.
- Added unit test `tests/unit/user-leagues-route.test.ts` to verify `min_submissions_per_day` and `max_submissions_per_day` are returned in the response payload.

## How to Test
1. Run `pnpm run test -- tests/unit/user-leagues-route.test.ts`.
2. Make a `GET /api/user/leagues` request with an authenticated user and verify that each league item in the `leagues` response array contains `max_submissions_per_day` and `min_submissions_per_day`.

## Checklist


blackhathckr (changes requested):
This can't merge — the two columns it selects don't exist. Merging it takes down league loading across the entire app.

### Blocker — `leagues.min/max_submissions_per_day` were dropped on Jul 13

`supabase/migrations/20260713_per_activity_limits.sql`:

```sql
-- 2. Drop the global league limits (added in 20260624_add_league_submission_limits.sql)
ALTER TABLE leagues
DROP CONSTRAINT IF EXISTS max_gte_min_submissions,
DROP CONSTRAINT IF EXISTS min_submissions_positive,
DROP COLUMN IF EXISTS min_submissions_per_day,
DROP COLUMN IF EXISTS max_submissions_per_day;
```

The global league-level limits were deliberately replaced by per-activity limits on `leagueactivities` (`min_frequency`, `frequency`, `frequency_type`, `submission_window_days`). No later migration re-adds them.

Verified against all three live databases just now:

```
dev  → {"code":"42703","message":"column leagues.max_submissions_per_day does not exist"}
qa   → {"code":"42703","message":"column leagues.max_submissions_per_day does not exist"}
prod → {"code":"42703","message":"column leagues.max_submissions_per_day does not exist"}
```

This PR adds those column names to four `leagues` SELECT strings. PostgREST rejects the whole query with a 42703, so each of these breaks outright:

- `src/app/api/user/leagues/route.ts:94` — **the endpoint #717 is about**
- `src/lib/services/leagues/league-query-service.ts:26` (`getLeagueById`)
- `src/lib/services/leagues/league-query-service.ts:81` (`getLeaguesForUser`)
- `src/lib/services/leagues/league-query-service.ts:144` (`getUserLeaguesWithRoles`)

Those last three are on nearly every authenticated page. And `updateLeague` now writes both columns into `allowedUpdates`, so any league settings save that includes them fails too.

### Blocker 2 — the new test asserts the broken behaviour

`tests/unit/user-leagues-route.test.ts` mocks `getSupabaseServiceRole` wholesale, so no query ever reaches Postgres. Worse, it pins the bug in place:

```ts
select: (fields: string) => {
  expect(fields).toContain('min_submissions_per_day');
  expect(fields).toContain('max_submissions_per_day');
```

It asserts that the select string contains the two non-existent columns, then hands back a hand-written row containing them. It passes green while the endpoint 500s against every real database. A test for a "field missing from a select" bug has to exercise the real column list — otherwise it only confirms the string was edited.

### The underlying issue needs a decision before code

#717 assumes *"this is likely just adding it to the `/api/user/leagues` select/serializer for parity."* That premise is wrong — the field is absent because it was **removed**, not because it was forgotten. @dikshaa-012's finding that it's missing on every league on dev is exactly consistent with the drop.

So there's a product question to settle first, and it isn't ours to assume:

1. **Re-introduce a league-level daily cap** — needs a migration re-adding both columns with their CHECK constraints, a host-facing setting to configure it, and enforcement wired in; or
2. **Mobile consumes the per-activity model instead** — `leagueactivities.min_frequency` / `frequency` / `frequency_type`, which is what web moved to in #713. Then #717 gets reframed as "mobile's daily-limit feature targets a removed field" and mfl-app#229 changes rather than the API.

Please settle this with @dikshaa-012 (mfl-app#229 depends on the answer) and update #717 before rewriting the PR. Note also that "min submissions per day" and "max submissions per day" are not the same concept as `min_frequency`, so option 2 isn't a rename.

### Worth doing while you're here

The reason the schema looked live is that develop still carries dead references to the dropped columns. These are pre-existing, not from this PR, but they're actively misleading and one of them is a latent failure:

- `src/app/api/leagues/[id]/route.ts:74-75` — zod validators for both fields on the league PATCH body
- `src/lib/services/entries-mutations.ts:28` — `leagues!inner ( max_submissions_per_day )`. If `EntryMutationService.createEntry` is ever called it errors, hits `if (member.error) { console.error('Invalid league member'); return null; }`, and silently drops the entry with a misleading log.
- `src/lib/services/leagues/league-types.ts:19-20` — both fields on `LeagueInput`

Cleaning these up would be a genuinely useful PR on its own.

Requesting changes. Happy to re-review once the direction on #717 is settled.

PR #833, opened 2026-08-01
feat(challenges): separate proof description from challenge descripti…

GitGuru-sudo:
## Summary

Adds a dedicated `proof_description` to challenges so proof instructions are a first-class field instead of being crammed into the challenge description, and surfaces it on all three player-facing surfaces.

## Related Issue

Closes #750

## ⚠️ Migration required before deploy

`supabase/migrations/20260801_proof_description.sql` must be applied **before** this code reaches any environment. The GET route selects `proof_description` on both `specialchallenges` and `leagueschallenges`; without the column PostgREST returns `42703` and the challenges page fails outright.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [x] UI/UX improvement
- [ ] Documentation

## Changes Made

**Schema.** `proof_description` added to both `specialchallenges` and `leagueschallenges` — customs have no template row, so the column has to exist on both. Copied from template to challenge at launch, reusing the `#547` copy-at-launch idiom.

**Three surfaces.** The player card shows the proof text under "Proof required"; tapping the challenge title opens the full explanation (description + proof + rules); the submit-proof dialog shows it where proof is actually produced.

**Description split.** The proof sentence is stripped out of No Sugar's template description with a `LIKE`-guarded update, so host edits are never clobbered.

### Review follow-ups

- **Removed `'No SNA Challenge'` from the seed.** It is not a name variant of No Sugar — it is a separate #545 template, `proof_mechanism = 'honour'` per `20260706` line 50, so it collects no proof at all. Verified on dev: `No SNA Challenge` is `honour` with `proof_required` null, `No Sugar` is `photo` with the certificate wording. Seeding it would have told players to produce a certificate for a challenge with no proof flow.
- **Removed the commented-out `HUMAN-GATED` block.** A migration cannot ship with an unresolved "fill before merge" TODO. Templates whose wording is not yet written are left NULL — which renders nothing rather than anything invented — and named in a comment.
- **Seed from `rules->>'proof_required'` for every template including No Sugar**, rather than restating the sentence. `20260728` already stored it, and the restatement already differed by a prefix, so the two copies would have drifted on the next copy edit. Whitespace is collapsed on the way in because the stored value is line-wrapped.
- **Proof precedence now applies on all three surfaces.** `view-rules-dialog` had `proof_description ?? rules.proof_required`, but the card and submit dialog read the column only — so a challenge holding its text in the jsonb showed proof in the details dialog and nothing where it is submitted. Extracted `resolveProofText()` and call it from all three, with unit tests.
- **`ViewRulesDialog` retitled** from `{title} - Rules` to `{title} - Details`, since it now leads with description and proof and is opened from the challenge title as the full explanation.

## How to Test

Apply the migration first (this repo uses **pnpm**, and migrations are applied manually per environment).

On dev the migration populates three templates from their existing `rules.proof_required` — Testing, No Sugar, Summer. Everything else stays NULL.

1. Open a league's challenges list. A challenge with proof text shows it under "Proof required" on the card.
2. Tap the challenge title — the details dialog opens with Description, Proof and Rules, titled "… - Details".
3. Tap Submit Proof — the same proof text appears above the upload field.
4. Check a challenge whose proof text lives only in `rules.proof_required` (not the new column): it must appear on **all three** surfaces, not just the details dialog.
5. Check `No SNA Challenge`: it is honour-based and must show **no** proof block anywhere.

## Screenshots / Recordings

Player card, details dialog and submit dialog all gained a proof block — worth a look.

## Checklist


blackhathckr (dismissed):
Solid shape for #750 — putting the column on **both** `specialchallenges` and `leagueschallenges` is the right call (customs have no template row), the LIKE-guarded description split is the careful way to do it, and the `#547` copy-at-launch idiom is reused correctly. The three surfaces (card / detail dialog / submit dialog) match what the issue asked for.

Blockers below are in the migration.

### Blocker 1 — `'No SNA Challenge'` is a different template and is being given the wrong proof text

```sql
UPDATE public.specialchallenges SET
  proof_description = 'A signed certificate from your spouse (or parent/child if no spouse), or a screenshot of their message certifying you went sugar-free.'
WHERE name IN ('No Sugar', 'No SNA Challenge')
```

The PR body calls these "both name variants". They aren't the same challenge:

- `'No Sugar'` is the ready-made template seeded in `20260728_no_sugar_ready_made_template.sql` (#776), `proof_mechanism = 'photo'`.
- `'No SNA Challenge'` is a separate client/PRD (#545) template, marked `lifecycle_stage = 'default'` in `20260705_add_challenge_lifecycle_stage.sql:51` and given its own attributes in `20260706_challenge_template_completeness.sql:50` — where **`proof_mechanism = 'honour'`** and `expected_benefits = 'Builds awareness, self-control, accountability'`.

So this writes "submit a signed certificate from your spouse" onto an **honour-based** challenge that by design collects no proof at all. Players get told to produce a certificate for a challenge that has no proof flow.

Drop `'No SNA Challenge'` from that `IN` list. If it needs its own proof text, it belongs in the human-gated block below with wording appropriate to an honour challenge.

### Blocker 2 — the migration ships with an unresolved pre-merge TODO

```sql
-- 5. HUMAN-GATED: remaining templates — fill from Nalini's doc before merge. Never fabricate.
-- UPDATE public.specialchallenges SET proof_description = '...' WHERE name = 'Try New Challenge';
-- UPDATE public.specialchallenges SET proof_description = '...' WHERE name = 'Experience Sharing';
-- UPDATE public.specialchallenges SET proof_description = '...' WHERE name = 'Workout Bingo';
```

The comment says "before merge" and it hasn't happened. Not fabricating the text was the right instinct, but this can't merge in limbo — the ready-made templates the issue names would ship with an empty proof block while the feature claims they're pre-filled.

Either fill them from the doc in this PR, or cut the commented-out block, state in the PR body which templates are intentionally deferred, and open a follow-up issue to seed them. Don't leave commented-out SQL in a shipped migration.

### Must fix — proof precedence is applied on only one of the three surfaces

`view-rules-dialog.tsx` establishes the precedence:

```ts
// #750 precedence: dedicated column wins, jsonb key as fallback
const proofText = proofDescription ?? rules?.proof_required ?? null;
```

But the other two surfaces read the column only:

- `challenge-card.tsx` — `{challenge.proof_description && ( ... "Proof required" ... )}`
- `submit-proof-dialog.tsx` — `{challenge?.proof_description && ( ... "What counts as valid proof:" ... )}`

Both receive the full `challenge` object, and the GET route already returns `rules`. Any challenge holding its proof text in `rules.proof_required` but not in the new column shows the proof block in the detail dialog and **nothing on the card or in the submit dialog** — which is the one place the issue most cares about, since that's where proof is actually submitted. Apply the same `?? rules?.proof_required` fallback in both, or pull the precedence into one small helper and call it from all three.

### Nits (all must be fixed)

**1. Two copies of the No Sugar proof sentence that will drift.** `20260728_no_sugar_ready_made_template.sql:36` already sets `rules.proof_required` to *"One of: a signed certificate from your spouse (or parent/child if no spouse), or a screenshot of their message certifying you went sugar-free."* Step 2 now writes a near-identical but not identical string into `proof_description`. Same sentence, two columns, already differing by a prefix — the next copy edit will update one and not the other. Either seed `proof_description` from `rules->>'proof_required'` for this row (step 3 already does exactly that for every other template) or clear `rules.proof_required` once the column is populated.

**2. `ViewRulesDialog` title is now wrong.** It still renders `{title} - Rules`, but the dialog now leads with Challenge Description and Proof, with rules third — and the card opens it from the challenge *title*, which per #750 is "the full detailed explanation". Change to `{title} - Details` (or just the challenge name).

**3. PR body: `npm run db:push`.** This repo is pnpm.

Address the two blockers, the precedence gap, and the three nits, then re-request review.

PR #828, opened 2026-07-30
fix(challenges): gate Submit Proof to the submission window (#748)

GitGuru-sudo:
## Summary
  Gates the "Submit Proof" button on player challenge cards to the challenge's submission window, and rejects
  out-of-window submissions server-side as defense in depth. Completion-type challenges open for proof only from
  `end_date` (end..end+1, the #743 weight-loss shape); continuous challenges keep opening at `start_date` — until #747's
  per-type defaults land, every type is wired as continuous, so behaviour is unchanged today.

  ## Related Issue
  Closes #748
  
  ## Type of Change
  - [x] Bug fix
  - [ ] New feature
  - [ ] Refactor (no functional change)
  - [ ] UI/UX improvement
  - [ ] Chore (dependencies, config, CI)
  - [ ] Documentation

  ## Changes Made
  - **`src/lib/utils/challenge-logic.ts`** — new pure helper `getSubmissionWindow({start_date, end_date,
  submission_opens_on?, is_completion_type?, now?, toleranceDays?})` returning `{opensOn, closesOn, isOpen}`;
  calendar-day comparison reusing `toCalendarDay` from weight-loss scoring (#743 pattern). `canSubmitChallenge` now
  takes `windowOpen` and returns `false` before the window opens.
  - **`src/components/challenges/challenge-card.tsx`** — card computes the window, hides Submit Proof while it's closed,
  and shows a "Submissions open <date>" state line instead (own-flow types `weight_loss`/`tournament` excluded — they
  never use the Submit Proof flow).
  - **`src/app/api/leagues/[id]/challenges/[challengeId]/submissions/route.ts`** — POST rejects out-of-window
  submissions with a 400 and a clear message; `toleranceDays=1` so UTC hosting never rejects a legitimately in-window
  local submit (#743 contract); resubmissions stay exempt (same as the existing late cutoff).
  - **`src/lib/services/challenges/weight-loss-scoring.ts`** — export `toCalendarDay` (was private) so the window helper
  reuses it instead of duplicating date parsing; no behaviour change.
  - **`src/lib/utils/__tests__/challenge-logic.test.ts`** — new `getSubmissionWindow` cases (continuous opens on start;
  completion opens on end, closed before, open end..end+1, closed after; tolerance widens edges; null dates ⇒ no
  constraint) + `windowOpen: false` gating case; existing `canSubmitChallenge` cases updated with `windowOpen: true`
  (behaviour-identical).
[...]

PR #834, opened 2026-08-01
fix(ai): exclude trial-period effort entries from AI context once lea…

GitGuru-sudo:
## Summary

Adds a conditional `start_date` clamp to every AI-facing `effortentry` query so trial-period entries stop feeding AI context once a league has officially started. During the trial window no filter is applied, so trial entries still appear and users get the full experience before start day. Also pages three league-wide fetches that were silently truncating at PostgREST's 1000-row cap.

## Related Issue

Closes #565

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] UI/UX improvement
- [ ] Documentation

### Trial-window clamp (#565 proper)

The issue listed 3 query paths; a grep of `.from('effortentry')` across the AI services surfaced 10 clamp points across 6 files, all covered:

- `src/app/api/leagues/[id]/ai-context/route.ts` — member entries query (365-row window)
- `src/lib/services/ai-insights.ts` — `getAiContext` member entries query (500-row window)
- `src/lib/services/ai/ai-coach-context.ts` — recent-entries (7d) and all-entries (30d) queries in `buildPersonalContext`, plus the team-rank aggregation
- `src/lib/services/ai/coach-context-builder.ts` — `getPlayerContexts` and `getInactivityGap`
- `src/lib/services/ai/health-context-builder.ts` — recent (14d) and all-entries fetches
- `src/lib/services/ai/player-insight-service.ts` — recent-entries and inactivity-gap queries

Supporting changes:

- **New shared helper** `src/lib/services/ai/trial-window.ts`. The `isLeagueStarted` computation and the object-or-array `leagues` embed unwrap were repeated in all six files; both now live in one place, with unit tests.
- **Calendar-day comparison.** `start_date` is a bare DATE column, so `new Date() >= new Date(start_date)` compared a local timestamp against UTC midnight and evaluated `false` for the first 5.5 hours of start day in IST — leaking trial entries into AI context exactly on the day it matters. The helper compares calendar days via `toCalendarDay`, the same approach `challenge-logic.ts` uses for challenge submission windows.
- **`challenge_submissions` clamped** in `health-context-builder`, which #565 explicitly asks for. This filters on `created_at`, **not** `submission_date`: per the `20260524` migration, `submission_date` is only populated for `is_unique_workout` challenges and is null for every row on dev, so clamping on it would have zeroed the count.

Notes on scope:

- The clamp is applied **before** the `.limit()` in each case, so a newest-N window can't be filled entirely with trial rows.
- Where a rolling window already exists (7d/14d/30d), the clamp composes with it — two `gte` filters on `date` AND together to the later bound.
- Three `effortentry` queries were deliberately **not** touched because they are already scoped to a single day via `.eq('date', today)` and cannot return trial rows: `ai-context/route.ts:82`, `ai-insights.ts:51`, `health-context-builder.ts:116`.

### Row-cap truncation

PostgREST caps a response at 1000 rows. Three queries here fan out over every member in a league with no pagination, so they were returning a silently truncated set — no error, just missing rows and wrong totals. `leaderboard-logic.ts` already pages around this by hand; `src/lib/services/ai/paged-fetch.ts` makes the same guarantee reusable.

- `ai-coach-context.ts` team rank — restored to the embedded `effortentry(...)` select with the clamp applied to the embed (`.gte('effortentry.date', …)`). The embed returns one row per member, so the cap is unreachable.
- `health-context-builder.ts` — both fetches now paged.
- `coach-context-builder.ts` `getPlayerContexts` — now paged.

Measured on a dev fixture (45 members, 25 days in, 1170 post-start entries): the unpaged query returned exactly **1000 of 1170** rows, and `HealthContextBuilder` reported team totals of `[338, 338, 324]` summing to 1000 instead of `[390, 390, 390]` summing to 1170, under-counting 7 of 45 members in the AI host digest.

## How to Test

Fixture league on dev: `BACKLOG-565-TEST` (`07b94e70-4ec2-4a06-8895-baef5cdb8ba7`), `start_date` 2026-07-08, 45 members across 3 teams, 1395 entries — 225 dated before `start_date` at RR 5, 1170 on/after at RR 1. Any trial leak inflates totals visibly. All fixture users log in with password `Test@1234`.

1. As `backlog-565-u01@test.local`, open `/leagues/07b94e70-4ec2-4a06-8895-baef5cdb8ba7/ai-coach`. Total points should read **26.0** and team standing **390 pts**. A total of 51 means trial entries leaked; a standing below 390 means the row cap is truncating.
2. As `backlog-565-u00@test.local` (host), open `/leagues/{id}/ai-manager`. All three teams should show **390** points.
3. Set the league's `start_date` to a future date to simulate the trial window and reload the AI Coach. The total should rise to **51** — during trial the clamp must not apply. Restore `start_date` afterwards.
4. Confirm team rank in the AI Coach matches the leaderboard's team ordering for the same league.

## Screenshots / Recordings

Not applicable — no UI change.

## Checklist


blackhathckr (dismissed):
Good coverage of the audit — you went beyond the 3 paths #565 listed and clamped 7 sites across 6 files. I verified the three effortentry queries you *didn't* touch (`ai-context/route.ts:82`, `ai-insights.ts:51`, `health-context-builder.ts:116`) are all `.eq('date', today)`-scoped, so they're correctly exempt. `start_date` is in the league select on all three entry points, so the flag actually evaluates.

Two blockers.

### Blocker 1 — PR body is the wrong PR's body

The entire description is copy-pasted from #830 (Ready-made terminology). It says **`Closes #749`**, so merging this auto-closes the terminology issue and leaves **#565 open**. The title, the diff, and every code comment say #565.

Rewrite the body for this PR: summary, `Closes #565`, the real file list, and real test steps.

### Blocker 2 — `ai-coach-context.ts` team-rank rewrite silently truncates at 1000 rows

The comment on the rewrite says:

> `// #565: site 7 — the nested effortentry embed can't take a conditional filter`

That's not correct — PostgREST filters embedded resources, and this repo already relies on it in ~10 places (`.eq('leaguemembers.team_id', teamId)` in `entries-queries.ts:54`, `.eq('activities.activity_name', ...)` in `entries/upsert/route.ts`, `.eq('challenge_subteams.league_challenge_id', ...)` in the submissions route). The one-line fix on the original query was:

```ts
let q = supabase
  .from('leaguemembers')
  .select('team_id, effortentry(rr_value, status)')
  .eq('league_id', leagueId);
if (isLeagueStarted && leagueStartDate) {
  q = q.gte('effortentry.date', leagueStartDate);   // filters the embedded rows
}
```

Because that wasn't done, the query was restructured into a flat fetch over every member in the league:

```ts
let teamEntriesQuery = supabase
  .from('effortentry')
  .select('league_member_id, rr_value, status')
  .in('league_member_id', teamLmIds);
// ...
const { data: teamEntries } = teamLmIds.length > 0 ? await teamEntriesQuery : { data: [] };
```

No `.range()`, no `.limit()`, no pagination. The old embed returned **one row per member** (top-level row count = member count, always small). The new one returns **one row per entry for the whole league** — and PostgREST caps that. This repo knows it does: `src/lib/services/leaderboard-logic.ts:319` paginates the exact same table with `PAGE_SIZE = 1000` and a `.range()` loop specifically because of this cap.

A 40-member league at 30 days is already past 1000 entries. Past that point `teamPoints` is computed from a truncated set, so `teamRankStr` reports a wrong point total and a wrong rank — to the AI Coach, silently, with no error. FFL2-scale leagues are well past this today.

Fix by filtering the embed as above (preferred — keeps the row shape, drops the `.in()` entirely), or paginate with the `leaderboard-logic.ts` `PAGE_SIZE` loop.

### Nits (all must be fixed)

**1. `new Date() >= new Date(start_date)` is a UTC-vs-local comparison.** `start_date` is a DATE column, so `new Date('2026-08-01')` is `2026-08-01T00:00:00Z`. For an IST user opening the AI Coach at 02:00 IST on start day (= 20:30 UTC on the previous day), `isLeagueStarted` is `false` and trial entries leak into AI context until 05:30 IST. Compare calendar days as strings instead — these files already build `new Date().toISOString().split('T')[0]` — or reuse `toCalendarDay` from `weight-loss-scoring.ts` (exported in #828).

**2. The flag block is copy-pasted 7 times.** The same 6-line `isLeagueStarted` computation appears in all 6 files, and the `Array.isArray(x.leagues) ? x.leagues[0]?.start_date : x.leagues?.start_date` unwrap appears 4 times. Extract both into one shared helper so the timezone fix in nit 1 lands in a single place instead of seven.

**3. `challenge_submissions` not addressed.** #565 explicitly asks: *"Also check `challenge_submissions` queries in your AI files — if they aggregate challenge data, the same conditional filter may be needed."* `health-context-builder.ts:201` counts submissions per active challenge with no date bound. Either apply the clamp or state in the PR body why it's exempt — right now it's just unmentioned (because the body belongs to another PR).

Fix both blockers and the three nits, then re-request review.

PR #835, opened 2026-08-02
hotfix: auto rest days stopped Jul 27 — ambiguous leaguemembers→users embed (PGRST201)

blackhathckr:
## Summary

Production hotfix. Auto rest days have not been assigned to **any** league since **Jul 27**, and five live API paths have been returning errors, all from the same root cause: an ambiguous PostgREST embed between `leaguemembers` and `users`.

## Root cause

`leaguemembers` has **three** foreign keys to `users`:

- `leaguemembers_user_id_fkey`
- `leaguemembers_created_by_fkey`
- `leaguemembers_modified_by_fkey`

A bare `users ( ... )` embed can't be resolved, so PostgREST rejects the entire query:

```
HTTP 300  PGRST201
"Could not embed because more than one relationship was found for 'leaguemembers' and 'users'"
```

### How it broke the cron

PR #520 (`feat/timezone-cron`) changed the auto-rest-day members query to fetch each member's timezone via `users ( last_known_timezone )`. The handler treats a query error as a per-league failure:

```ts
if (membersError) {
  console.error(`Error fetching members for league ${league.league_id}:`, membersError);
  continue;   // skips the entire league
}
```

So every league was skipped, every night, and the job still returned `{ success: true, assigned: 0 }`.

**Deploy timeline** (from GitHub deployment records) — the commit is dated Jul 22 but only reached production on Jul 27:

| production deploy | cron run | result |
| --- | --- | --- |
| `04d85dd5` — Jul 24 | Jul 26 08:30 UTC | 12 rest days assigned |
| `3332de17` — **Jul 27 05:53** | Jul 27 08:30 UTC | **0** |
| — | Jul 28 → Aug 1 | **0** every night |

No production deploy occurred between Jul 24 and Jul 27, so the last good run was on the old code and the first failed run was hours after the new code shipped. Platform-wide: **126 rest days assigned in the 11 days before, 0 in the 7 days since.**

## Changes

Every fix pins the relationship to `leaguemembers_user_id_fkey`, per PostgREST's own hint. Each was verified against production — the existing select returns **300**, the pinned select returns **200**.

| file | impact before fix |
| --- | --- |
| `api/cron/auto-rest-day/route.ts` | every league skipped; no rest days assigned since Jul 27 |
| `api/leagues/[id]/ai-context/route.ts` | member query returned null, so the route returned **404 "League or membership not found"** for every user |
| `api/leagues/[id]/engagement/route.ts` | two occurrences |
| `api/leagues/[id]/partner-activity/route.ts` | member list query failed |
| `api/leagues/[id]/balance/route.ts` | member list query failed |
| `lib/services/sponsors/sponsor-analytics-collector.ts` | aliased `user:users(...)`, same failure |

## Not included

`src/lib/services/memberships.ts` and `src/lib/membership.ts` carry the same ambiguous embed **and** select `role`, `age` and `gender` — none of which exist on `leaguemembers`. Actual columns:

```
league_member_id, user_id, league_id, team_id, created_by, created_date,
modified_by, modified_date, suspicious_proof_strikes,
suspicious_proof_last_strike_at, league_profile_picture_url, role_id
```

That needs more than an embed fix, so it's deliberately out of scope for a hotfix and will be raised separately. `membership.ts` has no importers.

## Follow-up needed after deploy

1. The cron only ever processes **yesterday**, so **Jul 26–31 will not self-heal** and needs a chronological backfill (assigning a rest day changes eligibility for the following day, so it can't be a naive per-day pass).
2. The job returns `{ success: true, assigned: 0 }` even when every league errors. It should surface a failed-league count so a silent zero can't look like success for seven days again.

## How to Test

1. Confirm the ambiguous select still fails and the pinned one succeeds:
   - `leaguemembers?select=league_member_id,user_id,users(last_known_timezone)` → 300
   - `leaguemembers?select=league_member_id,user_id,users!leaguemembers_user_id_fkey(last_known_timezone)` → 200
2. After deploy, invoke `GET /api/cron/auto-rest-day` with the `CRON_SECRET` bearer token and confirm `assigned > 0`.
3. Open a league dashboard as a player and confirm `/api/leagues/[id]/ai-context` returns 200 rather than 404.
4. `npx jest src/app/api/cron/auto-rest-day/route.test.ts` — 3/3 pass.

## Checklist

- [x] Verified against the production database
- [x] Existing cron unit tests pass
- [x] No behaviour change beyond resolving the relationship
- [x] Scope limited to the ambiguous-embed failures


adminmfl (approved):
Approved for production hotfix.

Root cause is confirmed against the production database rather than inferred: the ambiguous embed returns `300 / PGRST201` and the FK-pinned version returns `200`, for every one of the six call sites. The deploy timeline from GitHub's deployment records places the break precisely at the Jul 27 production deploy, with no deploy between Jul 24 and Jul 27 — so the last good cron run was on the old code and the first failed run was hours after the new code shipped.

Fix is minimal and mechanical: pin `leaguemembers_user_id_fkey`, no behaviour change beyond making the queries resolve. Correctly leaves `memberships.ts` out, since that additionally selects `role`/`age`/`gender` which don't exist on the table and needs a real fix rather than a hotfix.

Merging.

PR #863, opened 2026-08-03
hotfix: restore AI standings context and align points/run rate across screens

blackhathckr:
## Summary

Hotfix for four issues reported from the **live FFL 2 league**. Every one was reproduced against the production database before any code changed. Branched from `main`, to be backmerged `main → staging → qa → develop`.

## 1. AI coach told a first-place team it had zero points

`src/lib/services/ai-insights.ts` held:

```ts
// For now, let's keep it minimal to fix the 5.3s LCP.
const leaderboardData = null;
```

Left behind when the leaderboard HTTP call was removed for performance. Everything downstream reads from it, so `teamPoints = 0`, `teamRank = null`, `playerPoints = 0`, `pointsBehindLeader = 0`. The model was handed `Team Points: 0` and wrote *"Start strong, Fitness Champs — zero points means fresh potential to climb!"* for a team on **288 points at rank #1**.

It also silently degraded `/api/leagues/[id]/ai-motivate`.

Now calls `calculateLeaderboard` **directly**. That keeps the original latency win — no HTTP hop, no auth round-trip, no JSON serialisation — while restoring real numbers, and degrades to empty on error rather than taking the whole context down.

`playerPoints` also moved to `total_points`; quoting `points` back at a player understates their score by their challenge points.

## 2. My Team showed fewer points than the leaderboard

`my-team-view/page.tsx` and `my-team/page.tsx` both mapped `i.points` — activity points only — so every player's challenge score was missing.

Two of the client's screenshots are **one minute apart**:

| | Leaderboard 5:25 | My Team 5:26 |
|---|---|---|
| Ajay | **29** | **27** |
| Ananth | 28 | 28 |

The gap is exactly the 2 points from Ajay's approved *No Sugar* submission. Ananth has no challenge submission and matches on both. Both pages now use `total_points`; the request is unfiltered (`full=true`), so that is activity + challenge.

## 3. Report and My Zone disagreed on run rate

**1.59** on the League Summary versus **1.46** in My Zone, for the same player, from identical data.

The report recomputed run rate over `type === 'workout'` entries, dropping rest days — and the leaderboard counts a rest day at its `rr_value` of 1.0. That player had **6 rest days**.

Reproduced exactly against production, as of the screenshot timestamp:

| Method | n | Average | Screenshot |
|---|---|---|---|
| Report (workouts only) | 20 | 1.5924 → **1.59** | 1.59 ✅ |
| Leaderboard (rest days included) | 26 | 1.4557 → **1.46** | 1.46 ✅ |

The report already calls `calculateLeaderboard` for rankings (#787), so it now takes `avg_rr` from that same row — equal **by construction** rather than by two implementations happening to agree. The local average survives only as a fallback for a player with no ranked entry.

> Correcting one detail from my initial triage: I first said the report also counted rejected entries. It does not — the collector filters to `approved`. Rest-day exclusion is the entire difference.

## 4. Captains invited to click a player, but nothing happened

The roster row's `onClick` is gated on `player_team_workout_visibility`, but the hint *"Click any player to view their submissions"* rendered unconditionally. FFL2 had the flag **off**, so captains were told to tap a player and nothing happened. The hint is now behind the same flag as the handler.

(`my-team/page.tsx` rows are unconditionally clickable, so its wording is accurate and unchanged.)

## Applied directly to the production database — no deploy needed

Two of the reported issues were league configuration, not code. Applied and verified, with prior values backed up:

- **`submission_window_days = 1`** on all 14 FFL2 activities — every one was `NULL`, which collapses the date picker to today only. Both the client form and the server check in `entries/upsert` honour this symmetrically, so **logging yesterday works now**.
- **`player_team_workout_visibility = true`** on FFL2 — captains can open a player's submissions.

## Not fixed here — flagging

A player whose entry is rejected still cannot re-log that date from **Log Activity**; the reupload lives in **My Submissions**. The flow works (39 rejections, 33 reuploads completed) but it isn't where players look. Worth a follow-up issue rather than a change in a live-league hotfix.

Also observed, currently harmless: all 4 FFL2 teams have `teams.league_id = NULL`, 6 members have no team, and all 7 challenge submissions have `team_id` / `submission_date` NULL. Scoring reads the team from `leaguemembers`, so nothing is broken today.

## Verification

- New test fails **3 assertions** on unfixed `main` and passes here — it genuinely guards the bug
- Full suite **335 passing**; the 8 failing suites are a pre-existing missing `vitest` dependency, present on `main` too
- `tsc --noEmit` at the **93-error `main` baseline**, none in touched files
- `pnpm build` passes

## Migration

**None.**

## Checklist


adminmfl (approved):
Approved for production. Root causes verified against the live database, config fixes already applied and confirmed, and the new test fails on unfixed main.

PR #867, opened 2026-08-03
feat(challenges): let a host record a challenge submission for a player

blackhathckr:
## Summary

Reported from the live FFL 2 league: a player had valid challenge proof but could not upload it, and **the host had no way to add it for them**. Review only ever edited submissions that already existed, so with no row there was nothing to edit and no way to create one — the participant simply went unscored.

## What changed

`POST /api/leagues/[id]/challenges/[challengeId]/submissions` now accepts an optional **`leagueMemberId`**.

- Honoured **only** for a host or governor.
- The target must belong to **this** league — verified before anything is written, so a host cannot attach a submission to a member of a league they don't run.
- A player who sends the field for someone else is **refused (403)**, not quietly submitted for themselves.

Two gates are relaxed on this path, both deliberate and host-only:

| Gate | Why it's relaxed |
|---|---|
| Submission window | A missing proof is almost always noticed while reviewing results — i.e. *after* submissions close. Enforcing the window would leave the host at the same dead end. The host is already the approval authority, so the window isn't protecting anything from them. |
| `proofUrl` required | The host has usually been *shown* the evidence rather than handed a file. They can attach it afterwards through the existing edit. |

Neither relaxation applies to players.

**Points are unchanged** — they still come from the challenge total and remain capped by it, so this cannot award more than the challenge is worth.

## UI

The Review dialog now lists members with no submission for that challenge and records one against the selected player. It only renders when there is at least one such member.

## Already handled on production

The specific player the client asked about (Hustle & Muscle) has been given their 2 points directly on the production database, so they are not left waiting on this deploy. Verified: **28 activity + 2 challenge = 30**.

## Verification

- `tsc --noEmit` at the **93-error `main` baseline**, none in touched files
- Reviewed against the live challenge's real state: No Sugar ended Aug 1, so `effectiveStatus` is `submission_closed` — exactly the case the window relaxation exists for

## Migration

**None.**

## Checklist


adminmfl (approved):
Approved. Host-only, league-membership verified, and the points cap is unchanged.

PR #877, opened 2026-08-03
fix(leagues): validate numeric fields before create and surface real …

GitGuru-sudo:
## Summary
Fix Host league creation when persisted numeric form values are blank: parsing blank values produced `NaN`, which serialised as JSON `null` and could produce an unclear create failure. The client now validates the required team count and rest days before submission, sends the derived participant estimate, and displays the API's specific error message.

## Related Issue
Closes #849

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] UI/UX improvement
- [ ] Documentation

## Changes Made
- `src/app/(app)/leagues/create/page.tsx` — validate `num_teams` (minimum 2) and `rest_days` (0 or more) before building the create payload; send the existing derived `estimatedParticipants` value rather than parsing a non-existent participant-count input; surface the server error message instead of replacing it with a generic banner.
- `src/app/api/leagues/route.ts` — optional numeric inputs accept `null` as absent and normalise it to `undefined`; Zod validation responses include the first field-level error message.

## How to Test
1. Run `pnpm dev`, sign in as **Host**, and open **/leagues/create**.
2. Fill the league name, dates/duration, and Free tier. Leave **Number of Teams** empty, then select Create League. Expect the inline error **Please enter the number of teams (minimum 2)** and no POST to `/api/leagues`.
3. Enter `4` teams and a valid rest-day value, then create the league. Expect the success dialog and confetti; its player count should use the derived team estimate (`20` players for 4 teams).
4. Verify validation: `1` team is blocked; a blank or negative rest-day value is blocked.
5. Confirm a server-side validation error is shown in the page banner rather than a generic **Failed to create league** message.
6. Confirm the Paid tier still opens Razorpay, cancellation shows **Payment was cancelled**, and the Quick Start wizard is unaffected.

## Screenshots / Recordings

Screenshot 2026-08-03 220510
Screenshot 2026-08-03 220723

## Checklist
- [ ] My branch is up to date with `develop`


blackhathckr (changes requested):
Good diagnosis, and the root cause in the issue is exactly right — `parseInt('')` → `NaN` → `JSON.stringify` → `null`, and `.optional()` permits a missing key but not an explicit null. I traced the fix end to end and the free-tier path does now surface the real message: the create branch throws `new Error(createData.error || …)`, which reaches the outer catch you changed, so `num_teams: Expected number…` lands in the banner rather than the generic string.

The server relaxation is also safe — I checked `league-mutation-service.ts:39-40`, where a missing value falls back to `num_teams: data.num_teams || 4` and `rest_days: data.rest_days ?? 1`. So treating null as absent can't create a broken league.

Three things before this can merge.

### 1. Blocker — `max_participants` is now mandatory, but the page is built to treat it as optional

The new guard blocks submit when it is empty:

```ts
const maxParticipants = parseInt(formData.max_participants);
if (Number.isNaN(maxParticipants) || maxParticipants < numTeams) {
  setError('Please enter max participants (at least one per team)');
  return;
}
```

But the rest of this same file deliberately defaults it. Twice:

```ts
// page.tsx:245 and :269
const estimatedParticipants =
  parseInt(formData.max_participants) || parseInt(formData.num_teams) * 5;
```

So "blank" was a supported state with a documented fallback of `num_teams × 5`, used for tier recommendation and the price preview. This change turns a host who leaves it blank — which the pricing code expects — into a host who cannot create a league at all.

Worse, the value is **never stored**. It is not a column on `leagues`; I checked against production and `select=max_participants` returns **400**. `createLeague` doesn't reference it either. It only affects which tier is recommended and what the host is quoted.

So this now blocks creation on a field that changes nothing on the league record. Please either drop the hard guard and keep the existing `|| num_teams * 5` fallback, or — if the intent is that it should be stored — say so and we treat persisting it as its own piece of work. Blocking on a phantom field is the one option I don't think we should ship.

(`num_teams` and `rest_days` are a different case: both are persisted, both genuinely needed the guard.)

### 2. Blocker — the PR doesn't link its issue

The body says "No GitHub issue yet", but **#849** exists, is assigned to you, and its Root Cause / Fix sections are precisely this change. Please put `Closes #849` in the description — otherwise the issue stays open after merge and the next person re-investigates a fixed bug.

The `Refactor/bug-league-create-num-teams.md` the body references also isn't in the diff (0 `.md` files). Either include it or drop the reference.

### 3. Nit — the checklist says the branch is up to date; it is 18 commits behind

```
$ git rev-list --left-right --count pr877...origin/develop
1   18
```

Please rebase. `develop` has moved a fair amount, including scoring changes that landed today.

### Nit — formatting churn is mixed in with the fix

`route.ts` carries a set of prettier-only reflows (trailing commas, wrapped ternaries) alongside the real change. They're harmless, but they pad the diff and make the substantive edit harder to pick out. Worth keeping formatting-only passes in their own commit.

---

Changes requested. The core fix is correct and well-reasoned — it's the `max_participants` guard and the missing issue link that need resolving.

blackhathckr (changes requested):
Both blockers from the last round are cleared, and cleanly:

- The `max_participants` guard is gone and `|| num_teams * 5` is back, matching what `:239` and `:263` already do.
- `Closes #849` is in the body.
- Branch is now ahead-only (`1 → 0` behind).

The server relaxation still checks out — `createLeague` (`league-mutation-service.ts:39-40`) falls back to `num_teams: data.num_teams || 4` / `rest_days: data.rest_days ?? 1`, so null-as-absent can't produce a broken league.

One new blocker, which this PR is the first change to make reachable.

### Blocker — the success dialog will read "NaN Players"

`page.tsx:620`, in the `step === 'success'` screen:

```tsx
<p className="num text-[22px] font-extrabold …">
  {parseInt(formData.max_participants)}
</p>
<p …>Players</p>
```

`formData.max_participants` is **never assigned**. It's initialised to `''` at `:89`, and the only writers are:

- `:349` — the generic `handleChange`, which needs an input carrying `name="max_participants"`. There is no such input anywhere in the file; grepping `participants` in the JSX returns only the computed usages.
- `:203` — the `source_league_id` prefill, which sets `league_name`, `num_teams`, `rest_days`, `rr_formula`, `is_public`, `is_exclusive`. Not `max_participants`.

So it is `''` for every host, on every path, and `parseInt('')` is `NaN`. React renders that as the literal text `NaN`.

This didn't show before because the dialog was unreachable: `step === 'success'` is only set after the POST succeeds, and pre-#849 **every** create from this page 400'd on exactly this field. Your fix is what opens the path — the first host to create a league on this branch gets a celebration modal reading `NaN Players` next to their team count.

Use the memo that's already there:

```tsx
{estimatedParticipants}
```

`:356-362` computes precisely this and already handles the NaN case (`return Number.isNaN(numTeams) ? 0 : numTeams * 5`). It's also what `:721` passes down, so the dialog would finally agree with the tier card.

### Nit — the payload re-derives what the memo already holds

```ts
// :441-443
max_participants:
  parseInt(formData.max_participants) || parseInt(formData.num_teams) * 5,
```

That's now the fourth copy of this expression (`:239`, `:263`, the `:356` memo, here). `estimatedParticipants` is in scope. Worth noting the two forms aren't identical either — the memo requires `> 0`, the `||` form treats `0` as falsy and lands in the same place by accident rather than by rule. One source is better than four that agree by coincidence.

Separately: `max_participants` is not a column on `leagues` and `createLeague` never reads it, so this key is discarded server-side. Not something to fix here, but it argues for the memo rather than a fresh parse.

### Nit — PR body still documents the removed guard

Two places describe behaviour that isn't in the diff any more:

- **Changes Made** — "validate … `max_participants` (>= teams)".
- **How to Test** step 4 — "Teams `4`, participants `2` → blocked (at least one per team)". A reviewer following that will find it *not* blocked, which is now correct, and reasonably conclude the fix is broken.

Please bring both in line with what shipped.

### Nit — formatting churn is still mixed in

`route.ts` still carries the prettier-only reflows (trailing commas, wrapped ternaries) alongside the schema change. Same note as last round: fine in isolation, but it pads a 2-file diff and buries the substantive edit.

---

Changes requested. The `NaN` is the only one that reaches a user; the rest is tidy-up. Fix that and the body, and this is ready.

blackhathckr (changes requested):
`3a90580` fixes both code points, and cleanly.

- `page.tsx:618` is now `{estimatedParticipants}` — the success dialog can no longer render `NaN`. The memo at `:356` returns `Number.isNaN(numTeams) ? 0 : numTeams * 5`, so the blank case resolves to a number in every path.
- `page.tsx:441` uses the same memo for the payload, so the dialog, the tier card (`:719`) and the wire value can't drift apart.

I checked the ordering too — memo at `:356`, `handleCreateLeague` at `:384`, so it's in scope. Merge is clean against `develop`.

The code is done. What's left is the description, which still describes a guard that was removed two rounds ago.

### Nit — "Changes Made" documents validation that isn't in the diff

> validate `num_teams` (min 2), **`max_participants` (>= teams)**, `rest_days` (>= 0)

That middle guard was removed in `bbac187` at my request, because `max_participants` is not a column on `leagues` and blocking creation on it blocked a phantom field. The line has survived three rounds unchanged.

### Nit — "How to Test" steps 3 and 4 can't be followed

Both reference a participants input that does not exist on this page:

> 3. Enter `4` teams, `20` participants → Create League.
> 4. Teams `4`, participants `2` → blocked (at least one per team)

```
$ grep -n 'name="max_participants"\|id="max_participants"\|Max Participants' page.tsx
(no matches)
```

There is no such field in the JSX. `formData.max_participants` is initialised to `''` at `:89` and never written — the only writers are the generic `handleChange` (which needs an input that isn't there) and the `source_league_id` prefill (which doesn't set it). That's exactly why the dialog was rendering `NaN`.

Step 4 is the one that matters: someone following it will find the submit **not** blocked, which is now correct behaviour, and reasonably file it as a regression.

### Nit — checklist says up to date; branch is 19 behind

```
$ git rev-list --left-right --count <branch>...origin/develop
3   19
```

That's my hotfix merges from today, not your doing, and `git merge-tree` comes back with **0 conflicts** — so nothing is blocked. But the box is ticked and the statement isn't true.

### Nit — formatting churn is still bundled in

Third time on this one: `route.ts` still carries the prettier-only reflows (wrapped chains, trailing commas) alongside the schema change. Harmless, but it turns a two-line schema edit into a 40-line diff.

---

Changes requested on the description. The implementation is correct and I'd merge it today — I'm not going to approve a PR whose test plan tells the next person to verify a guard we deliberately removed.

Also worth knowing for later, not for this PR: `:239` and `:263` still inline `parseInt(max_participants) || num_teams * 5` rather than using the memo. Those predate you and behave identically while the field is unreachable, so leave them — but if a `max_participants` input is ever added, the memo's `> 0` test and the `||` fallback will disagree on `0`.

blackhathckr (changes requested):
All four items are cleared, and one of them was cleared by doing more than I asked — thank you for splitting the formatting into `a94b3450`. I checked it: 39 insertions, 17 deletions, all prettier reflow, no logic. That is exactly the right shape.

Also verified:

- **Branch is `4 → 0`.** Rebased, no longer behind.
- **Body matches the diff.** "Changes Made" no longer describes the removed guard.
- **How to Test is executable.** Step 3 now checks the derived count (20 players for 4 teams) instead of a field that isn't filled.

---

## First — a correction to my own review

I twice told you there is **no `max_participants` input**, and used that as the justification for two separate points. That was wrong, and you should know before you act on the rest of it.

The input exists:

```
src/components/league/league-form-section.tsx:308-317
  <Label htmlFor="max_participants" …>
  <Input id="max_participants" name="max_participants" value={formData.max_participants} …>
```

I grepped only `create/page.tsx` and concluded the field was unreachable. It's rendered by `LeagueFormSection` and written through `handleFormChange`, along with `num_teams` and `rest_days` — all three live in the same child component. My grep was too narrow and I stated the conclusion with more confidence than the evidence supported.

**What this does not change:** the fix is still correct, and so is the `NaN` blocker.

`max_participants` is **optional**, so it is frequently left blank — and blank is what produced `parseInt('') === NaN` in the success dialog. The bug was real; my explanation of *why* the field was empty was not. `estimatedParticipants` handles both cases properly: it returns the entered value when one is present (`:358`, guarded on `> 0`) and falls back to `num_teams * 5` when it isn't. That is the right fix either way.

**What this does change — one line in your body:**

> send the existing derived `estimatedParticipants` value rather than parsing a **non-existent participant-count input**

That phrasing came from my review and it is inaccurate. The input exists; it is simply optional and often empty. Please reword — something like "rather than parsing an optional field that is frequently blank".

**And one thing worth a second look, which I got partly wrong too.** I justified dropping the `max_participants >= num_teams` guard on the grounds that the field was a phantom. It isn't. The better justification — which still holds — is that `createLeague` never persists it (`league-mutation-service.ts:36-42` writes `num_teams` and `rest_days` only), so validating it blocks creation over a value the server discards.

But it is not entirely inert: it feeds tier recommendation (`:240`) and the pricing call (`:264`, `:303`). So a host who enters 3 participants for 4 teams now gets a pricing estimate built on a nonsensical number, with nothing telling them.

I am not asking you to restore the guard in this PR — it is out of scope and the field is not persisted. Please raise it as a separate issue so the decision is recorded rather than lost.

---

## Nit — the screenshots predate the fix

Both attachments are from **3 August, 22:05 and 22:07**. The commit that replaced `parseInt(formData.max_participants)` with `estimatedParticipants` in the success dialog is `4ba84eae`, and the branch has since been rebased. So the images predate the change by a week.

Step 3 now asks a verifier to confirm the dialog shows 20 players for 4 teams. There is no screenshot of that, and if either existing image shows the success dialog it shows the pre-fix value.

Retake the success-dialog screenshot against the current branch.

While you are in there: the Screenshots section still carries the template comment (`<!-- If UI changed, add before/after screenshots… -->`) above the images. Delete it.

---

Changes requested on the body wording and the screenshot. The code is correct and I have no further concerns with it.

PR #880, opened 2026-08-04
fix(challenges): allow a 0% to 0% weight-loss tier for holding steady

blackhathckr:
## The blocker

The host could not launch the FFL2 weight-loss challenge. Their announced scheme pays **1 point for finishing at the same weight**, which needs a tier from `0%` to `0%` — and the validator rejected it:

> Each non-final tier needs a max % above its min %.

The tier editor **offers** that row, so the host filled it in, was refused, and the challenge sat unlaunched *after the announcement had already gone out to players*.

This is a code defect, not a misconfiguration. I reproduced both of the host's attempts against `main`'s real compiled functions and got the exact error string from their screenshots.

## Why relaxing the validator alone would have been worse

`tierForPct` returns `null` for `pctLoss <= 0` **before any tier is read**, and the range test is `[min, max)` — so `p >= 0 && p < 0` can never be true. A `0–0` tier could never match anything.

Verified: with only the validator relaxed, `pointsForPct(0, clientTiers)` still returns **0**. The challenge would have launched and then silently scored 0 for exactly the players it was meant to reward — surfacing at the awards ceremony rather than at launch. So both halves change together.

## What changed

**1. `validateTiers`** — a no-change tier is allowed at 0 and nowhere else. A zero-width tier at 3% could never match (nobody loses exactly 3.0000%), so permitting one would just relocate the silent-zero trap. A second no-change tier is rejected.

**2. `tierForPct`** — exactly 0% matches **only** when the host has explicitly configured a no-change tier. Weight gain still returns `null` unconditionally. A no-change tier is skipped in the range loop so it cannot swallow a real loss.

**3. Sort tie-break** — both rows start at `min_pct = 0`, so without this, validity depended on which row the host typed first. I hit this while testing.

**4. Two display strings** that would otherwise contradict the score — the participant tier list hardcoded *"Same weight or heavier → 0 pts"*, and `predict()` told a level player they had earned nothing.

## The host's exact configuration

| | on `main` | with this fix |
|---|---|---|
| `0–0=1, 0–1=2, 1–2=3, 2–3=4, 3–4=5, 4–5=6, 5+=7` | **rejected** | **valid** |

Every band pays what was announced:

| % change | announced | this fix |
|---|---|---|
| gain | 0 | 0 |
| 0% | 1 | **1** |
| 0.5% | 2 | 2 |
| 1% | 3 | 3 |
| 2% | 4 | 4 |
| 3% | 5 | 5 |
| 4% | 6 | 6 |
| 5%+ | 7 | 7 |

## Existing leagues are untouched

This is the part that mattered most. `DEFAULT_TIERS` spans `[0, 5)`. Had the 0-guard simply been dropped, **every level player in every live league would have started scoring 2 points**.

Scoping the new behaviour to an explicitly configured tier means output is **identical** to production across the range:

| % | prod | this fix |
|---|---|---|
| −1 | 0 | 0 |
| 0 | 0 | 0 |
| 0.01 | 2 | 2 |
| 4.9 | 2 | 2 |
| 5 | 3 | 3 |
| 10 | 4 | 4 |
| 25 | 4 | 4 |

Nothing else in weight loss is affected — logging, `computePctLoss`, the progress graph, and `maxTierPoints` (still 7, so `total_points` is unchanged) all use unmodified paths.

## Verification

- **44/44** in the weight-loss suite, with every pre-existing test unchanged and passing
- **348 passing** overall; the 8 failing suites are the pre-existing missing `vitest` dependency, identical on `main`
- `tsc --noEmit` at **100 = main's exact baseline**, none in touched files

New tests cover: the announced scheme validating, 0% scoring its configured points, gain still scoring 0, a real loss not being swallowed, both row orderings, a duplicate no-change tier rejected, a zero-width tier away from 0 rejected, and — explicitly — that existing leagues still score 0 at 0%.

## Migration

**None.**

## Checklist


adminmfl (approved):
Approved for production. Verified against the host's exact configuration, and existing leagues score identically to main across the range.

PR #884, opened 2026-08-04
feat(challenges): require proof for weight-loss initial and final weights

blackhathckr:
## Summary

Weight was the only scoring submission on the platform logged purely on trust. The FFL2 host asked for proof mid-challenge, so players now attach a screenshot certifying the number.

Also renames **"Base weight" → "Initial weight"** throughout the UI (labels only — `log_type='start'` is unchanged, so no data migration).

## Migration — **already applied to production**

```sql
ALTER TABLE public.challenge_weight_logs
  ADD COLUMN IF NOT EXISTS proof_url text;
```

Nullable and idempotent. **No new storage bucket** — `challenge-proofs` already exists and `.env.prod` already points at it.

## Proof rules

| Log type | Proof |
|---|---|
| Initial | **required** |
| Final | **required** |
| Progress | not required — feeds the graph only, scores nothing |

Enforced in the API, not by a `NOT NULL`, so logs written before this shipped keep working. Those players get a **PATCH** route to attach proof to an existing log rather than being stuck with an unverifiable number every one of their points is computed from. That route only ever sets proof, only on the caller's own log, and only when the log has none.

## Two defects found while testing — both older than the proof work

**1. The host could not see who had logged.** The results table was built from `challenge_submissions`, which only exist once a player logs a **final** weight. So for the entire challenge the host saw *"No results yet"* — precisely when they need to see who has logged an initial weight and chase the people who have not.

It now derives from the logs. The column heading is **"Latest"** rather than "End", since mid-challenge that value is a progress entry and "End" implied the player had finished; Points reads **"Not final yet"** instead of `0`, which had looked like they scored nothing.

**2. A host who is also a player was stuck on "your score is being calculated".** This route branches on *membership* role while the page picks its view from the client's *active* role. A host viewing as Player got the host payload — everyone's `results`, but no personal `result` or `prediction` — so the participant view fell back to the calculating message while their submission sat **scored and approved** in the database. Both payloads now carry the caller's own result and prediction.

An existing comment on `endDate` shows this same mismatch was hit before and fixed for one field only.

## Also in here

- **Live preview contradicted the score.** It computed points correctly and then discarded them when the change was `<= 0`, hardcoding *"scores 0 pts"*. With a no-change tier configured (`0%→0% = 2 pts`) it told a player they had earned nothing while the server credited them. It now branches on what the tiers actually award. Verified against a real prod config: 76kg → 0, **75kg → 2**, 74kg → 3, 67kg → 4.
- **Upload accepted GIF, the bucket rejects it** — the failure landed at storage as a raw error instead of a clear message. Aligned to the bucket (JPG/PNG/WebP).
- **Swap button hidden** on the host's request. The dialog, handler and API route are left intact, so restoring it is one line.

## UX

Upload shows a dashed drop area → spinner with *"Uploading your proof…"* → a **thumbnail of the actual image** with Replace, so the player can see the right photo attached before committing a weight. Proofs open in the existing shadcn `ViewProofDialog`, not a new tab.

On the final days both actions share one weight box, so each now sits in its own bordered group with a one-line description, and a short note explains that Update progress adds a graph point while Log final weight scores. Proof lives inside the final group and is no longer sent — or consumed — by a progress log.

## Verification

- **348 tests passing**; the 8 failing suites are the pre-existing missing `vitest` dependency, identical on `main`
- `tsc --noEmit` at **93 — main's exact baseline**, none in touched files
- Exercised end to end against the production database on a live weight-loss challenge: upload → store → read back, host table, retroactive proof, and scoring all confirmed

## Checklist


adminmfl (approved):
Approved. Migration already applied to production; verified end to end against the live weight-loss challenge.

PR #888, opened 2026-08-04
fix: one scoring source for My Zone, and make rejections actionable

blackhathckr:
## Summary

Fixes the items raised in the Aug 3 review. Every finding below was reproduced against **production data** before changing anything — including two that turned out **not** to be bugs.

## Root cause: scoring was implemented twice

The leaderboard and analytics use `calculateLeaderboard`. The dashboard behind My Zone had its **own parallel maths** and never called it (`calculateLeaderboard`: **0** occurrences). That copy diverged three ways at once — no team-size normalisation, no #752 guest filter, and challenge points left out of the personal total.

## #new — Team rank 1 vs 2 (**proven**)

FFL2 has `normalize_points_by_team_size: true` with uneven teams (11/8/8/7). Computed from live data:

| | Rank 1 | Rank 2 |
|---|---|---|
| **Raw** (old dashboard) | Vibe Fitness 300 | **Fitness Champs 219** |
| **Normalised** (leaderboard) | **Fitness Champs 301** | Hustle & Muscle 300 |

Exactly the reported mismatch.

## #2 — My Zone understated points (**proven**)

`totalPoints` was activity-only. **All nine players with challenge points were short by exactly their challenge score:**

| Player | Leaderboard | My Zone | Diff |
|---|---|---|---|
| Satish, Ajay, Renuka, Kumudini, Sushma, Suvida, Deepya | 31 | 29 | −2 |
| Sunitha | 22 | 20 | −2 |
| Satyapriya | 22 | 21 | −1 |

The dashboard now reads standings from the engine, keeping the local computation only as a fallback if that call fails.

> Note: I could **not** reproduce a points mismatch on the **analytics** screen — it already reads `total_points` from the same engine. The client's two screenshots were ~12 hours apart with data changing between them. The confirmed mismatch is My Zone.

## #2/#3 — Analytics timezone (**real, intermittent**)

Analytics called the engine with **no timezone**, falling back to the server clock. For an IST league between **00:00 and 05:30**, that is the previous calendar day — and "today" bounds the scoring window. Now threaded through and part of the cache key, since viewers in different timezones legitimately get different answers.

I tested it: **no difference outside that window**, which is why it wasn't visible in the screenshots.

## #3 — The phantom rest day (**proven**)

Analytics *invents* rest days: a closed day left unlogged is silently reclassified as rest when allowance remains. That is precisely *"it's showing as a rest day used, which is not the case."*

The behaviour is deliberate — it mirrors what the auto-rest cron will do — so it stays, but the projected count is now returned and marked with `*` in the roster instead of being presented as fact.

## #4 and #7 — **not bugs**

- **RR 2.00 with a rest day is correct.** Satish's three rest days are all `rejected_permanent`, so scoring rightly excludes them; all 29 counted entries are workouts at 2.0.
- **The full-points rule is already implemented.** It was being fed the inferred rest days above — a symptom, not a separate rule.

## #8/#9/#10 + "where's my rejected entry" (**confirmed gap**)

The count was plain text with nothing to click, and **no part of the app told a player an entry needed redoing** — so rejected days quietly became missed days. There is now a persistent alert above the fold and the tile links to the submissions screen, where the reupload flow already worked.

## #1 — Host add-submission

Was a single button that posted immediately with no proof and no points. Now a real form: player, proof upload, and an optional points override **capped at the challenge total**.

## Verification

- New parity test **fails 3/5 on unfixed `main`**, passes 5/5 here — it genuinely guards the bug
- **353 tests passing**; the 8 failing suites are the pre-existing missing `vitest` dependency, identical on `main`
- `tsc --noEmit` clean relative to `main`'s baseline

## Migration

**None.**

## Checklist


adminmfl (approved):
Approved. Findings reproduced against production data; parity test fails on unfixed main.

PR #892, opened 2026-08-04
fix: point the rejection alert at the reupload form, not a host-only page

blackhathckr:
## The bug

The **"Fix now"** alert I shipped in #888 linked to `/leagues/[id]/submissions`. That page is **host/governor only**, so a player who tapped it landed on:

> **Access Restricted** — Only host or governor can view all league submissions.

…with no way forward. Reported with a screen recording; the clip ends on that error screen.

That's worse than the silent number it replaced, because it actively invited the tap.

## Why I couldn't just retarget the link

There is **no player-facing submissions page on web**. The route was deleted in `703c2e4d` ("removed My Activities page"), which left three things orphaned:

- `MySubmissionsTable` — rendered nowhere
- `ReuploadDialog` — rendered nowhere
- `GET /api/leagues/[id]/my-submissions` — no page consumes it

And `/submissions` still redirects to the deleted `/leagues/[id]/my-submissions`.

## What actually works

The reupload entry point lives inside the **activity timeline**: open the rejected day → **Reupload** → navigates to the submit form with `resubmit` / `date` / `type` params (`activity-timeline.tsx:86`).

It works — but the player has to know to scroll to the timeline, find the right day, open it, and spot the button inside a dialog. **That is what "resubmission not clear" means.**

## The fix

The alert now builds that same URL directly and skips the hunt. `submit/page.tsx:79-117` reads exactly these params, so this reuses the existing flow rather than adding a second one.

The dashboard returns the rejected entries with the fields the form needs, newest first, and the alert **names the date** — so the player knows which day to re-photograph before they tap.

**Permanent rejections are counted but not offered.** They can't be resubmitted, and linking one would dead-end the player again — the same mistake in a different place. An entry already superseded by an approved one on the same date drops out, so the alert disappears once there's genuinely nothing left to fix.

## Verification

- 5 new tests covering the deep-link payload, ordering, permanent-rejection exclusion, superseded entries, and the empty case
- **358 tests passing**; the 8 failing suites are the pre-existing missing `vitest` dependency, identical on `main`
- `tsc --noEmit` at **100 — main's exact baseline**

## Not addressed here

The orphaned `my-submissions` page/table/dialog. Whether to restore that screen or delete the dead components is a bigger call than a hotfix should make — worth its own issue.

## Migration

**None.**

## Checklist


adminmfl (approved):
Approved. Reproduced from the client's recording; deep-links into the existing reupload flow rather than adding a second one.

PR #896, opened 2026-08-04
feat(challenges): let a host replace the proof on an existing submission

blackhathckr:
## The request

> "Update — should be able to update proof for the existing ones too."

Review could change a submission's **score** but not its **proof**. A host handed better evidence after the fact — a clearer photo, or the right one when the player attached the wrong file — had no way to update the record, so the stored proof could disagree with the decision made from it.

## What changed

Each submission row now offers **Replace proof** beside *View Proof*, using the same uploader as the add-on-behalf form.

The upload is **staged, not saved immediately**:

- Pick a file → *"Proof ready — press Update to save it"*
- **View Proof** then shows the staged image, so the host can confirm it's the right file before committing
- **undo** clears a mis-picked file before it has overwritten anything
- **Update / Approve** commits it

That ordering matters: overwriting on file-pick would mean a wrong tap silently destroys the original proof with no way back.

Server side, `validate` sets `proof_url` **only when `proofUrl` is present in the body**, so an ordinary approve or reject leaves the existing proof untouched. An empty string clears it rather than storing `""`.

## Overlap to settle at promotion time

`qa`/`develop` already carry **#795**'s edit dialog, which exposes proof as a **URL text field**. That covers pasting a link but not attaching a photo — which is what was actually asked for. Once this reaches `qa` the two overlap, and the text field is the one worth dropping.

Flagging rather than doing it here: deleting a field from a dialog that isn't on production yet is a call for the promotion, not a hotfix.

## Verification

- `tsc --noEmit` at **100 — main's exact baseline**, none in touched files
- **358 tests passing**; the 8 failing suites are the pre-existing missing `vitest` dependency, identical on `main`

## Migration

**None.**

## Checklist


adminmfl (approved):
Approved. Staged upload with undo, and proof_url only written when explicitly provided.

PR #901, opened 2026-08-04
fix(uploads): stop a rejected image surfacing as an HTML parse error

blackhathckr:
## Error 1 — the HTML parse error

A player uploading proof saw:

> Failed to upload file: Unexpected token '<', "&lt;html&gt; &lt;h"... is not valid JSON

**The HTML wasn't our response.** It was inside the storage provider's error message, which `upload/proof/route.ts:78` concatenated straight into ours:

```ts
{ error: 'Failed to upload file: ' + uploadError.message }
```

So the payload was *valid JSON containing an HTML page* — which is why the client's existing content-type guard in `upload.service.ts` didn't catch it. That guard was already correct; it just never fired.

**Why storage rejected the file:** this route accepted **`image/gif`**, and the `proofs` bucket does not. Verified against production — `allowed_mime_types` is `['image/jpeg', 'image/jpg', 'image/png', 'image/webp']`. A GIF passed our check and failed at the bucket **every time**.

The same mismatch was fixed on `challenge-proof` in #884 and missed here.

**Now:** the allow-list matches the bucket, the rejection names the type it received, and storage errors are logged in full but never shown raw — size errors get a size message, HTML-shaped errors get a plain retry message, anything else is truncated to 200 chars.

## Error 2 — improved, not fixed

The resubmission date check said only that the dates had to match, naming **neither**:

> Resubmissions must use the original submission date

A player looking at a date field the form had **already locked** couldn't tell what was wrong — and diagnosing the report meant querying the row by hand. It now names both:

> This resubmission is for your **2026-08-01** entry, but the form sent **2026-08-03**. Reopen it from the rejected entry and try again.

**The underlying cause could not be reproduced.** Against production FFL2:

- Most recent rejection is **Aug 2** — nothing on Aug 3 or Aug 4
- **Zero** entries dated Aug 3 are rejected or reuploads
- **Zero** resubmissions attempted on Aug 4
- Seven other players have working parent→child reupload pairs

Both routes into that form (the "Fix now" alert and the timeline's Reupload) take the id and the date from the **same row**, so neither can produce the mismatch. The date isn't user-editable either — `date-selector.tsx:47` renders it as read-only text during a resubmission. The improved message will identify the source if it recurs.

## Verified locally against the production database

Full cycle on a live league as host + player: log activity → reject → **Fix now** → correct screen, correct locked date → resubmit → success. Normal JPG/PNG uploads unaffected by the narrowed allow-list.

- `tsc --noEmit` at **100 — main's baseline**
- **358 tests passing**; the 8 failing suites are the pre-existing missing `vitest` dependency

## Migration

**None.**

## Checklist


adminmfl (approved):
Approved. Verified end to end against the production database on a live league.

PR #907, opened 2026-08-05
fix(weight-loss): make the initial weight correctable, widen the final window

blackhathckr:
Hotfix off `main`. Weight-loss weigh-in windows, requested for a live league whose weight-loss challenge is currently running.

## The problem

The initial weigh-in was **one-shot**. A mistyped baseline returned `409 Baseline weight already logged` and could never be corrected. Because the result is a percentage *of* that number, a wrong baseline silently skews every point for the whole challenge, and the player had no way out.

## What changed

**Initial weight — now correctable, inside a window**

- Loggable and freely overwritable across `start .. start+5` (six days)
- A repeat log **updates the same row** instead of 409ing, matching what the final weight already did
- Proof stays mandatory on every version

Wide on purpose: the window exists for integrity (you cannot set a baseline after you have already lost weight), but a narrow one shuts people out of the challenge entirely, since a final weight cannot be scored without a baseline.

**Clamped so it can never outlive the challenge**

`initialWeightWindowCloseDay` caps the close at the challenge's own end date. A six-day window on a two-day challenge would leave the baseline editable days after results were final — and since a corrected baseline re-scores, that would let someone raise their starting weight afterwards and inflate their percentage lost. Challenges of six days or more are completely unaffected.

**Final weight — `end..end+1` → `end-1..end+1`**

Widened, not shifted. The day before was requested; the day after was kept because it is what the app already promises players on screen, and removing it mid-challenge would take away a day they are counting on.

**A corrected baseline re-scores**

Gated on a submission already existing. `scoreWeightLossMember` falls back to the latest *progress* log when there is no final one, so an unguarded call would mint an approved, points-bearing result mid-challenge for anyone who had merely been tracking their graph.

**Result view — one Edit per weight**

The result view is an early return, so the initial weight was previously unreachable there entirely — which is the state every player with a result is in. Each weight now carries its own inline `Edit` on its own row; the single header button is gone, since it could only ever mean one of the two.

## Verification

Tested against the production database on a non-live test league, covering a real re-score:

| Step | Stored result |
|---|---|
| Baseline | `start 75, end 63, 16%, 4 pts` |
| Final weight → 58 | `start 75, end 58, 22.67%, 4 pts` |
| Initial weight → 65 | `start 65, end 58, 10.77%, 4 pts` |

Each overwrite updated the existing row rather than inserting a second one, and `reviewed_at` advanced each time, confirming the re-score fired. Percentages and tier selection were recomputed independently and matched exactly.

- Unit tests: **67** in the weight-loss suite, covering both windows, the clamp, calendar-day behaviour and timezone tolerance
- Full suite: **381 passing**
- `tsc` and `eslint`: error counts identical to `main`, none in the changed files
- Production build compiles clean

## Notes

- **No migration.** Deployable as-is.
- Windows remain fixed defaults. Making them host-configurable is tracked separately in #906, which needs a new migration (the existing `submission_opens_on`/`submission_closes_on` pair is a single window; weight loss needs two).


adminmfl (approved):
Approved. Verified against the production database on a non-live test league: each overwrite updated the existing row rather than inserting a second, and the corrected baseline re-scored (start 75 -> 65 recomputed 22.67% -> 10.77%). Clamping the initial window to the challenge end closes the post-result baseline edit. No migration required.

PR #928, opened 2026-08-06
fix(weight-loss): proof dialog reopening, and chart showing other members' weigh-ins

blackhathckr:
Two client-reported production bugs on the weight-loss challenge, both reproduced against prod data before fixing.

## 1. The "Submit Proof" dialog kept reappearing

**Reported:** *"Keep getting this even after proof is entered."*

**Cause, traced end to end:**
- FFL (`dc07df54…`) has exactly **one** active challenge and it is `weight_loss` (`1fd34132…`) — confirmed in prod.
- `LogChallengeButton` deep-links to `?submit=<id>` whenever `active.length === 1`.
- The challenges page auto-open effect called `handleOpenSubmit(target)` **without checking the challenge type**, while `canSubmitChallenge` already excludes tournaments and weight loss because they submit on their own screens.
- Weight loss writes to `challenge_weight_logs`, not `challenge_submissions`, so that upload could never be recorded.
- `?submit=` stays in the URL, so it reopened on every visit — hence "keep getting this".

**Fix:** the own-flow rule is now an exported helper, `usesOwnSubmissionFlow`, shared by both sides. The button links straight to the challenge screen for those types, and the deep-link handler routes there instead of opening a dialog that cannot do anything.

## 2. The graph did not match the initial weight

**Reported:** *"Weight in graph vs weight in initial weight is different."*

**Cause:** the chart was plotting **every member's** weigh-ins, not the viewer's.

The page picks its view from the **client** role while this API branches on **membership** role, so a host viewing as Player renders the participant view against the *host* payload — and that payload returns `logs` unfiltered.

The five points on his chart are exactly the five rows in the table, across two members:

| member | log | weight | logged_at |
| --- | --- | --- | --- |
| `1333362c` | progress | 94.7 | Aug 4 05:54 | ← chart's blue "Start" dot |
| `1333362c` | progress | 94.8 | Aug 4 06:11 |
| `c6ebb3a4` | **start** | **74.7** | Aug 5 02:58 | ← the card's "Initial weight" |
| `c6ebb3a4` | progress | 74.7 | Aug 5 02:59 |
| `1333362c` | start | 96 | Aug 5 16:15 | ← chart's red "Latest" dot |

`startWeight` takes the first `'start'` in array order (74.7, correct), while the chart's first point is the earliest log of *any* type (94.7, another player's). Hence the mismatch.

**Fix:** the host branch now also returns `ownLogs` — already computed there for the prediction — and the participant view reads that when present. The player branch was always filtered server-side and is untouched.

**This was also a data leak:** one player's weights were visible to another on that screen.

## Not included — tracked on #906

His third request, host-configurable weigh-in windows, is logged in full on **#906** with the clarified spec. Deliberately excluded because:

- It needs the migration #906 was raised for — #747's `submission_opens_on`/`submission_closes_on` is a **single** window; weight loss needs two.
- His proposed `start+2` default would close FFL's initial window **today**, contradicting the "9th is fine" he confirmed in the same conversation.
- His `end-2 … end` removes the day *after* the end date, so FFL's final window would go 13–15 Aug → 12–14 Aug, costing a day players may be relying on.
- He has since added "challenge end date should be extendable, and deadlines adjusted" — a further feature, and one that re-scores, which is exactly the hole #907 closed.

No FFL change was needed after all: its initial window already closes Sun 9 Aug, which he confirmed is fine.

## Verification

Traced against production data, not inferred. `pnpm build` passes. `pnpm lint` unchanged at `main`'s baseline of 50 errors / 728 warnings — none introduced. (For reference, the UI revamp sitting on `develop` clears 48 of those.)

No scoring, submission or window logic changed. `#907`'s weigh-in windows, the end-date clamp and the submission-gated re-score are untouched.


adminmfl (approved):
Approved. Both fixes were reproduced against production data before being written: the deep-link chain that opened an unusable proof dialog, and the host payload leaking every member's weigh-ins into a personal chart. No scoring or window logic touched — #907's clamps and re-score gating are intact. Host-configurable windows correctly deferred to #906.

PR #932, opened 2026-08-07
hotfix: weight-loss points via generic submit, corrected-baseline chart, mobile dialog scroll

blackhathckr:
## Summary

Production hotfix for three defects reported against FFL 2 on 7 Aug, all in the weight-loss / manual-entry area.

**1. Two players were credited the full 7 points for the weight-loss challenge, nine days before it ends.**

Not a date bug. `POST /api/leagues/[id]/challenges/[challengeId]/submissions` accepts *any* challenge type and writes `status: 'approved'` with `awarded_points: total_points`. Weight loss is not scored there — it is scored by `scoreWeightLossMember` from a start + end weigh-in, as a percentage of the baseline, inside the `end-1 .. end+1` window. Nothing on the server stopped a photo-proof upload from being pointed at a weight-loss challenge.

The two bad rows are distinguishable from real ones:

| | the two bad rows | a genuine weight-loss row |
|---|---|---|
| `awarded_points` | 7 (= `total_points`) | derived from `pct_change` |
| `start_weight` / `end_weight` / `pct_change` | **all null** | all populated |
| `proof_url` | a photo | null |
| `reviewed_at` | null | set |

One of the two players had a `start` log and progress logs but no `end` log; the other had **no weight logs at all**. Neither could have been scored by the real path.

`usesOwnSubmissionFlow` already encoded the rule that weight loss and tournaments own their submission flow — but only in the UI. This adds the server-side half.

**2. Correcting the initial weight left the graph showing the old value.** The number on the card updated, the chart did not, so the two contradicted each other.

`#907` made the initial weight correctable, and the weight-log route stamps `logged_at = now()` on a correction. The chart sorted by `logged_at` and then painted the blue "Start" dot at **index 0**, so a corrected baseline slid down the series and the dot landed on whichever progress log happened to sort first.

Reproduced against the reported data — baseline 96 kg created 4 Aug 05:51 and corrected 5 Aug 16:15, with progress logs of 94.7 and 94.8 on 4 Aug:

```
before   Aug 4  94.7  <- painted "Start"      after   Aug 4    96  <- "Start"
         Aug 4  94.8                                  Aug 4  94.7
         Aug 5    96  (the real baseline)              Aug 4  94.8
         Aug 7  95.3  <- "Latest"                      Aug 7  95.3  <- "Latest"
```

Now ordered and labelled by `created_at`, which a correction leaves alone, and the endpoints are identified by `log_type` instead of by position so an edit can never repaint them.

**3. The manual-entry dialog opened past its own top on a phone**, with the header and first field off-screen and no way to scroll back to them.

`DialogContent` is centred with `-translate-y-1/2` and capped at `90vh`, and the box itself does not scroll — only the body between the fixed header and footer does. `vh` is measured against the viewport with the browser toolbars *retracted*, so on a handset `90vh` can exceed what is actually on screen; the excess splits above and below the centre line, and the half above the top is unreachable. Changed to `svh` (smallest viewport height, toolbars shown). On desktop there is no dynamic toolbar and `svh === vh`, so nothing changes there.

## Related Issues

Reported directly against production on 7 Aug; issues to follow.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [x] UI/UX improvement
- [ ] Documentation

## Changes Made

- `src/app/api/leagues/[id]/challenges/[challengeId]/submissions/route.ts` — reject `weight_loss` and `tournament` in the POST, before any write, reusing `usesOwnSubmissionFlow`.
- `src/app/api/leagues/[id]/challenges/[challengeId]/weight-log/route.ts` — return `created_at` from both GET selects.
- `src/components/challenges/weight-loss/participant-view.tsx` — order and label the chart by `created_at`; colour the endpoints by `log_type`. `created_at` is optional on `WeightLog`, so a cached response without it falls back to `logged_at`.
- `src/components/ui/dialog.tsx`, `src/components/ui/alert-dialog.tsx` — `max-h-[90vh]` → `max-h-[90svh]`.
- `tests/unit/challenge-submissions-own-flow-guard.test.ts` — new.

## Deliberately not included

The `#747`/`#748` submission-window enforcement that `develop` carries. Production is missing every column it depends on — `submission_opens_on`, `submission_closes_on`, `is_completion_type`, `launched_at` — so shipping it here would `42703` the route. It needs its own migration and promotion.

## How to Test

1. As a player, open a **weight-loss** challenge and try to submit through the photo-proof dialog. Expect `400 "This challenge is scored from its own screen, not from a proof upload."` and **no row** in `challenge_submissions`. Repeat for a tournament challenge.
2. Confirm an **individual** challenge still submits and scores normally — the guard must not widen.
3. Log a weight-loss baseline, add a progress log, then **correct the baseline**. The blue "Start" dot must sit on the corrected value and stay at the left of the series; the red "Latest" dot on the most recent non-baseline log. The chart must agree with the "Initial weight" card.
4. On a **phone**, as host, open Player Activities → Manual Entry → Add workout. The dialog header and the first field (Type) must be visible and the form scrollable to the top.
5. Spot-check a few other dialogs on desktop for unchanged height.

## Verification

| | `main` | this branch |
|---|---|---|
| `pnpm build` | — | pass |
| `tsc --noEmit` errors | 109 | 109 (0 in touched files) |
| jest tests passing | 381 | 384 |
| jest suites failing to load | 8 | 8 (pre-existing resolution errors) |

The new test fails without the guard and passes with it, and asserts that nothing reaches `challenge_submissions` — the row was the incident, not the error message.

## Follow-ups (not in this PR)

- **Data correction.** Two `challenge_submissions` rows on production still hold the incorrect 7 points. The derived score tables (`specialchallengeteamscore`, `specialchallengeindividualuserscore`) are empty for this league, so removing those two rows is the whole correction. Awaiting a decision from the league owner.
- **"Submission Deadline" wording** on the activity config reads "+N days after activity date", but the implementation is a rolling window anchored on today (`earliest = today − N`). Identical at +1, divergent above it.
- **Rest-days-used `*`** is explained only in a `title` tooltip, which does not exist on touch. Wants a visible legend.
- **A `typecheck` script.** There is no `tsc --noEmit` in `package.json` and no CI gate, which is why 109 type errors sit unnoticed on `main`.

## Checklist


adminmfl (approved):
Approved for production.

Root cause is correctly identified and the fix is at the right layer: `usesOwnSubmissionFlow` already stated the rule, this puts it on the server where it binds. The guard sits before any write, so a rejected request cannot leave a row behind — and the test asserts that rather than just the status code.

Scope is right for a hotfix. Holding back the #747/#748 window enforcement is the correct call given production is missing all four columns it reads.

Verified: build green, no new type errors, 381 → 384 tests.

PR #937, opened 2026-08-07
hotfix: show recorded duration/distance/steps in the submission review dialog

blackhathckr:
## Summary

A host opening a player's activity submission sees the activity type, the points and the proof image — but **not the duration, distance, steps or holes** the score was computed from. Those are the numbers a review is about.

Reported against a Gym entry scoring **2.00 RR**; the underlying `effortentry` has `duration: 103`. The host could only get at that by reading the minutes out of the proof screenshot.

## Root cause

Nothing was missing from the data layer:

- `MySubmission` has carried `duration`, `distance`, `steps`, `holes` all along (`use-my-submissions.ts:18-21`)
- `/api/leagues/[id]/my-submissions` selects all four (`route.ts:121-124`)
- `MetricItem` was already written **in this same file** to render exactly this, complete with a null guard

It was simply never called. `MetricItem` had **no consumers anywhere in the repo**:

```
$ git grep -n "MetricItem" -- src/
src/components/submissions/detail-dialog/submission-detail-metrics.tsx:7:export function MetricItem({
```

So this wires up a component that was written for the job and left unconnected.

## Related Issue

Reported directly against production on 7 Aug; issue to follow.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [x] UI/UX improvement
- [ ] Documentation

## Changes Made

- `src/components/submissions/detail-dialog/submission-detail-metrics.tsx` — render the four measurements under a **Recorded** heading, using the existing `MetricItem`.

Behaviour worth noting:
- `MetricItem` returns `null` on a null value, so an activity shows **only** the measurement it uses — a Gym entry shows duration, a Steps entry shows steps, neither shows the other.
- The whole block is skipped when all four are null, so rest days and completion-type entries render exactly as before.
- Gated on `isWorkout`, consistent with the Activity Type block above it.
- Steps go through `toLocaleString()`; a five-digit step count is unreadable otherwise.

## How to Test

1. As host, open **Player Activities → Review** and open any approved **Gym** or duration-based submission. Expect a **Recorded** section showing *Duration — N min*, and the value to match the proof.
2. Open a **Steps** submission — expect *Steps* with a thousands separator, and **no** empty Duration/Distance/Holes tiles.
3. Open a **rest day** — expect no Recorded section at all.
4. Confirm Activity Type, Points Earned, Custom Fields and Player Notes are unchanged.

## Verification

| | `main` | this branch |
|---|---|---|
| `pnpm build` | — | pass |
| `tsc --noEmit` errors | 109 | 109 (0 in touched file) |
| jest tests passing | 384 | 384 |
| jest suites failing to load | 8 | 8 (pre-existing) |

## Checklist


adminmfl (approved):
Approved for production.

Good diagnosis — the fix is wiring, not new code. `MetricItem` was written for exactly this and had no consumers anywhere in the repo, so the reviewer-facing gap was a missed connection rather than missing data.

The null-per-item guard is the right shape: an activity shows only the measurement it uses, and the outer condition keeps rest days untouched.

PR #941, opened 2026-08-07
hotfix: surface every reported field in the submission review dialog

blackhathckr:
## Summary

Follow-on to #937. That added **duration, distance, steps and holes** to the review dialog. Three more fields a player can report were still rendered nowhere:

- **outcome** — the Win/Loss/Draw style label, when the host has configured outcomes on the activity
- **hr_avg** — average heart rate
- **calories_burned**

Same shape of gap as before: all three are already selected by `/api/leagues/[id]/my-submissions` (`:133`, `:136`, `:137`) and typed on `MySubmission`. Nothing rendered them.

They are not hypothetical. On production right now:

| field | entries carrying a value |
|---|---|
| `calories_burned` | **180** |
| `hr_avg` | **121** |
| `outcome` | **33** |

A host reviewing any of those was comparing a proof screenshot against a figure the app already held and would not show.

## On the Steps case that prompted this

Already fixed by #937 — `MetricItem` renders `steps` with a thousands separator. The reported entry has `steps: 12831`, matching the *12,831* in the proof. No further change needed there.

## Related Issue

Reported directly against production on 7 Aug; issue to follow.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [x] UI/UX improvement
- [ ] Documentation

## Changes Made

- `src/components/submissions/detail-dialog/submission-detail-metrics.tsx` — render `outcome`, `hr_avg` and `calories_burned` alongside the existing four, in the same **Recorded** block.

Notes:
- `outcome` needs no lookup. The API resolves configured tiers by `o.label === entry.outcome`, so the stored value **is** the label and renders as written.
- The presence test is now a named `hasRecordedValues` listing each field explicitly. Deriving it from the object would have hidden the next omission; spelled out, adding a field to `MySubmission` without surfacing it here reads as a gap rather than passing unnoticed.
- `MetricItem`'s null guard is unchanged, so an activity still shows only what it actually collected — a Steps entry shows steps and nothing else.
- Calories go through `toLocaleString()`, consistent with steps.

## How to Test

1. As host, open **Player Activities → Review** on a submission with heart rate or calories recorded — expect *Avg heart rate — N bpm* and *Calories — N kcal* under **Recorded**, matching the proof.
2. Open a submission on an activity with **outcomes configured** — expect *Outcome* showing the configured label (e.g. Win).
3. Open a plain **Steps** submission — expect steps only, with no empty tiles for the fields it does not use.
4. Open a **rest day** — expect no Recorded section.

## Verification

| | `main` | this branch |
|---|---|---|
| `pnpm build` | — | pass |
| `tsc --noEmit` errors | 109 | 109 (0 in touched file) |
| jest tests passing | 384 | 384 |
| jest suites failing to load | 8 | 8 (pre-existing) |

## Checklist


adminmfl (approved):
Approved for production.

Right call to check the rest of the shape rather than just adding the one field that was reported. The prod counts make the case — 180 entries with calories and 121 with heart rate were being reviewed against a proof image while the app held the number.

Naming the presence test and listing the fields explicitly is the part worth keeping. A derived check would have made the next omission invisible, which is exactly how this set accumulated.

PR #964, opened 2026-08-09
Fix/update rules

dikshaa-012:
## Summary
The player-facing Rules page never rendered the host's free-text rules summary — only the Host-only "Manage Rules" page displayed it. Players had no way to see it at all.

## Related Issue
Closes #956 

## Type of Change
- [x] Bug fix

## Changes Made
- Added a "Summary" card to `src/app/(app)/leagues/[id]/rules/page.tsx` rendering `rulesData.rules_summary`, matching the block already used on `/rules/manage`, minus the Edit button.
- Placed it first on the page, above "Activity or challenge?" — it's the host's own description of how their league runs, so it should be the first thing a player reads.

## How to Test
1. As Host, go to **Rules → Manage Rules**, set a rules summary, and save.
2. As Player, open the same league's **Rules** page.
3. Confirm the summary text now appears under a "Summary" heading, at the top of the page, above "Activity or challenge?".
4. Confirm a league with no summary set still shows no Summary card.

## Screenshots / Recordings
Screenshot 2026-08-10 033801
Screenshot 2026-08-10 033916

## Checklist


blackhathckr (changes requested):
The diagnosis is right, and the way you got there is the part worth calling out.

You started on the caching hypothesis, went back, disproved it, and posted a correction on #956 saying so. The real cause — the player-facing `/rules` page never rendered `rules_summary` at all, so there was nothing to go stale — is exactly the kind of finding that only comes from tracing rather than assuming. I'd rather see a revised root cause than a confident wrong one.

The fix itself checks out:

- The block matches `/rules/manage:161-166` including `whitespace-pre-wrap`, and correctly drops the edit button that only belongs on the host's page.
- It's inside the `!hasRules` else-branch, and `hasRules` (`:84`) already counted `rules_summary` — so a league with a summary and no document was already passing the empty-state check and then rendering nothing. That lines up precisely with your diagnosis.

Three things, all small.

### Nit — the PR body doesn't describe half the diff

**Changes Made** lists only the Summary card. The diff also contains three edits you haven't mentioned, in a second file:

```
rules/route.ts   + export const dynamic = 'force-dynamic'
rules/route.ts   + { headers: { 'Cache-Control': 'no-store' } }
rules/page.tsx   + fetch(..., { cache: 'no-store' })
```

And the checklist ticks *"I have not modified any files outside the scope of this task"*, against a PR that modifies the API route. Whoever reviews this at release time reads the body, not the diff.

### Nit — the cache changes should come out

From your own comment on #956:

> Cache-control changes made earlier … are harmless and left in, but weren't the actual fix.

Half right. They aren't the fix, but "harmless" isn't quite true either:

- `export const dynamic = 'force-dynamic'` is **redundant** here. This handler awaits `params` and does an auth check, so it was never statically optimised — the directive changes nothing. It's used in 34 other routes so it isn't out of place stylistically, but it's dead weight added on a hypothesis you disproved. Note it also applies module-wide, so it covers the `DELETE` and `POST` handlers too, not just the GET you were debugging.
- `Cache-Control: no-store` on the response **does** have an effect. It prevents any browser or CDN caching of the rules payload, for every player, on every open — on a page whose content changes maybe twice in a league's lifetime.

Since caching was ruled out as the cause, none of this is buying anything. Please strip all three and keep the PR to the one-file change it actually is. If you think any is worth having on its own merits, that's a separate PR with its own reasoning.

### Nit — step 3 of How to Test points at the wrong place

> Confirm the summary text now appears under a "Summary" heading, **between "Activity or challenge?" and "Allowed activities."**

It doesn't. On the branch:

```
:113  Activity or challenge?
:194  Allowed activities
:207  Summary          <- here
:216  rules_doc_url
```

It renders *after* Allowed activities, not between. Someone verifying this will look in the wrong spot and report it as not working.

Which raises the question worth answering rather than just correcting the text: **should** it be there? The summary is the host's own description of how their league runs — arguably the first thing a player should read on the Rules page, above a generic activity list. Right now it sits third. Your call, but say which you intended and make the body match.

---

Changes requested on the three nits. The fix is correct and I'd take it as soon as the diff matches what the PR says it is.

blackhathckr (changes requested):
All three cleared, and the diff is now what the PR says it is.

- **Cache changes gone.** `git diff --stat` is one file, 13 insertions, 0 deletions. `route.ts` is untouched — no `force-dynamic`, no `Cache-Control: no-store`, no `cache: 'no-store'` on the fetch.
- **Body matches the diff**, including a line explaining the placement decision.
- **How to Test step 3 now describes where the card actually renders.**

You also took the placement question and answered it rather than just correcting the text — Summary at `:105`, above "Activity or challenge?" at `:124`. That's the right call and the reasoning in the body is the right reasoning: it's the host's own description of how their league runs, so it should be the first thing a player reads.

I checked the conditional too. `hasRules` (`:82`) already counted `rules_summary`, so a league with a summary and no document was passing the empty-state check at `:185` and then rendering nothing. The new block sits above that branch and is independently gated, so an empty league still shows the empty state and nothing else changes.

One thing left.

### Nit — the screenshots predate the move

Both attachments are the ones from the original revision:

```
Screenshot 2026-08-10 033801
Screenshot 2026-08-10 033916
```

Taken at 03:38 and 03:39. The commit that moved the card to the top of the page — `e9db575f` — landed at **17:30**, roughly fourteen hours later. Before that, the card rendered *below* "Allowed activities".

So the screenshots show the old position, while the body and step 3 both say it now sits at the top above "Activity or challenge?". Whoever verifies this at release time will compare the two and conclude one of them is wrong.

Retake both against the current branch and this is done.

---

Changes requested on the screenshots only. The code is correct and I'd merge it the moment the evidence matches it.

Minor, and not something to act on: the branch still carries `85c6405c` ("prevent stale cache") whose changes were later reverted, so `develop` will take a commit that adds the cache directives and another that removes them. Harmless with merge commits, and not worth a rebase for a three-commit branch — noting it only so it doesn't read as an accident later.

PR #965, opened 2026-08-10
hotfix: rest-day donation approvals, rest-day replace copy, sticky error toasts

blackhathckr:
## Summary

Three defects reported from production on 10 Aug.

### 1. Rest-day donations could never be approved

The chain routes a request to the donor's team captain first — but nothing ever showed captains a queue. `use-rest-day-donations.ts:61` gated the fetch on `['host','governor']` and `page.tsx` gated the tab the same way. Requests reached **"Captain Review"** and stopped there permanently.

**Six are in that state on FFL 2 right now**, the oldest for two days.

What makes this a pure wiring bug: the API had implemented captain approval all along (`[donationId]/route.ts` stage 1), and `ApprovalQueueTab` was already written with a captain variant ("Approved by You", no Final Approve button). Neither was ever reachable.

**Now:** captains see the queue for their own team, and their approval can no longer block. After a **24h review window** it is assumed and the request moves on to the host or governor, whose approval is what actually transfers the days.

The window is derived from `created_at`, not advanced by a cron. A missed run would strand requests in exactly the state this exists to prevent — the #839 failure mode this codebase already documents for challenge lifecycles.

Past the window the captain also loses standing to **reject**, so a host cannot approve something a captain then overturns.

### 2. Replacing a rest day said "Activity"

`formatEntrySummary` derived its label from `workout_type`, which is null on a rest day, so it fell through to the literal `'Activity'`. A player replacing a rest day was told they were replacing an *"Activity, 1.0 RR"* — neither what they logged nor what they were about to lose. Title, field label and summary now all reflect the entry type.

### 3. Error toasts never went away

`toast.ts:22` set `duration: Infinity` on errors. A validation error such as *"Please enter steps"* then sat on screen indefinitely: the player corrected the field, submitted successfully, and the original error was **still there over the top of the success dialog** — visible in the reported screenshots.

Errors now clear after 8s, and a success dismisses any error still showing. Tracked by id, so only the error is dismissed — a bare `dismiss()` would also take down loading toasts from an unrelated in-flight request. Repeated errors replace rather than stack.

## Type of Change

- [x] Bug fix
- [x] UI/UX improvement

## Changes Made

- `src/lib/utils/rest-day-donation-window.ts` — **new.** The 24h captain window, derived. Fails *open* on an unparseable timestamp: a request whose age we cannot establish should keep a reviewer able to act, not silently lose one.
- `.../rest-day-donations/[donationId]/route.ts` — captain approve/reject gated on the window; host/governor approve `pending` **or** `captain_approved` → `approved`.
- `.../rest-day-donations/_hooks/use-rest-day-donations.ts` — captains fetch the queue.
- `.../rest-day-donations/page.tsx` — captains see the tab, scoped to their own team and to in-window requests only (showing a lapsed row would offer a button that 403s).
- `.../submit/_components/submission-dialogs.tsx` — rest-day-aware copy.
- `src/lib/toast.ts` — finite error duration, dismiss-on-success, no stacking.
- `tests/unit/rest-day-donation-window.test.ts` — **new**, 13 cases.

## Verification against production data

Ran the shipped helper against the six live FFL 2 requests:

```
  id        age(h)  captainCanAct  approvalAssumed  hostSees
  c8e841fc     8.6           true            false  approvable
  b604669e    31.9          false             true  approvable
  16a7440e    42.7          false             true  approvable
  2d628a10    48.1          false             true  approvable
  345ff3b9    48.1          false             true  approvable
  1329fb6e    48.2          false             true  approvable
```

The 8.6h request is still its captain's to action; the five older than a day are assumed approved and waiting on the host. **All six remain approvable**, so closing the captain window strands nothing.

| | `main` | this branch |
|---|---|---|
| `pnpm build` | — | pass |
| `tsc --noEmit` errors | 101 | 101 (0 in touched files) |
| tests passing | 384 | **397** |
| suites failing to load | 8 | 8 (pre-existing) |

## Assumptions — flagged for review

The brief was *"Host/governor need to approve — captain approval is assumed after a day."* Two details it did not settle, decided as follows:

1. **Host/governor can approve immediately**, without waiting out the 24h. The captain step is optional; making the host wait for an optional step would be a worse dead end than the one being fixed.
2. **Once assumed, the captain can no longer reject.** Otherwise a host approves something the captain then overturns.

Say so and I'll change either.

## Not addressed — needs a product decision

Rest days currently move **the moment a request is made**. `getMemberFinalRestDays` sums over `['pending','captain_approved','approved']`, so the receiver can already spend days nobody has approved — all six open FFL 2 requests are already in effect.

That means the host approval this PR enforces happens *after* the thing it authorises. Making it real means counting only `approved` on the receiver side while keeping `pending` on the donor side (so nobody donates the same days twice). **Deliberately not changed here** — it alters live balances and needs sign-off.

## How to Test

1. As a **captain**, open Rest Day Donations → the **Approval Queue** tab is now present, showing your team's requests from the last 24h. Approve one → status becomes Captain Review complete, awaiting host.
2. Confirm a request older than 24h does **not** appear in the captain's queue.
3. As **host/governor**, confirm every open request is approvable regardless of age, and approving transfers the days.
4. Log a rest day, then submit a workout for the same date → the dialog reads **"Replace Rest Day?"** with *"Current rest day"*.
5. On Log Activity, submit with an empty required field → error appears; correct it and submit → the error is gone, not sitting behind the success dialog. Leave an error alone → it clears in 8s.

## Checklist


adminmfl (approved):
Approved for production.

The donation fix is at the right layer — the API and the queue component both already supported captains, so this is wiring plus the window rule rather than new behaviour bolted on. Deriving the 24h from `created_at` instead of adding a cron is the correct call and consistent with what we learned in #839.

Two things worth calling out as good practice: the helper fails *open* on an unparseable timestamp rather than silently removing a reviewer, and the captain's queue hides lapsed rows so the UI never offers a button the API will 403.

Verifying against the six live FFL 2 requests — showing that the 8.6h one stays with its captain while the five older ones move on, and that none are stranded — is the evidence that matters here.

Correctly held back the `pending`-counts-as-active change. That alters live balances and is the product owner's call.

PR #971, opened 2026-08-10
feat(weight-loss): widen initial-weight window to end-1; captain 'accepts', governor 'approves'

blackhathckr:
Two requests from the league owner.

## 1. Initial weight editable until the day before the challenge ends

Was `start..start+5` clamped to the end date. Now **`start..end-1`**, with `INITIAL_WEIGHT_WINDOW_DAYS` kept only as the fallback for a challenge with no end date.

For **FFL 2** (Aug 4 → Aug 14) the window moves from **Aug 4–9** to **Aug 4–13**.

Rationale: people join late, weigh on a different scale, or mistype — and a window that shut on day six left them stuck with a baseline that every point they score is a percentage of.

### Accepted trade-off, flagged before building

This **overlaps the final-weight window**, which opens at `end-1`. On that shared day a player can log a final weight, see the score, then raise their baseline — and `weight-log/route.ts:215-224` re-scores a corrected baseline whenever a result already exists:

```ts
let shouldRescore = logType === 'end';
if (logType === 'start' && existingLogId) {
  shouldRescore = Boolean(existingResult);   // ← baseline edit re-scores
}
```

So the higher percentage pays out. This was raised with the league owner, who chose the wider window over closing the gap. It is documented at the function and **pinned by a test**, so it reads as a decision rather than an oversight later. If it is ever exploited, the fix is to freeze the baseline once a final weight exists — not to narrow the window again.

### Awards day — no change needed

Logging a **final** weight on awards day already works. The window is `end-1..end+1` on the client and `end-2..end+2` on the server. FFL 2 ends **Aug 14**, the ceremony is **Aug 15** — already inside both. A configurable awards date for other leagues would need a new field and host UI; out of scope here.

## 2. Captain "accepts", Governor/Host "approves"

The captain's step is a team-level acknowledgement, not the decision that transfers the days — calling both "approve" made them look interchangeable.

| Where | Before | After |
|---|---|---|
| Captain's button | Approve | **Accept** |
| Captain's section | Approved by You | **Accepted by You** |
| Badge (host view) | Captain Approved | **Captain Accepted** |
| Governor button | Final Approve | **Approve** |
| Donor status | Captain Review | **Awaiting captain** |

Labels only — no state machine or permission change.

## Type of Change

- [x] Bug fix
- [x] UI/UX improvement

## Verification

| | `main` | this branch |
|---|---|---|
| `pnpm build` | — | pass |
| `tsc` source errors | 93 | **93** (diffed line by line — no new) |
| tests passing | 384 | **391** |

The initial-window suite was rewritten rather than extended: it asserted the old `start+5` rule throughout, so leaving it would have meant tests passing against behaviour we no longer have. 61 cases in that file now, including the deliberate overlap.

## How to Test

1. On FFL 2, as a player with a baseline, edit your initial weight on **Aug 10–13** — previously refused after Aug 9, now accepted.
2. Confirm it is **refused on Aug 14** (the final day) and after.
3. Log a final weight on **Aug 15** (awards day) — already worked, confirm unchanged.
4. As **captain**, open Rest Day Donations → Queue: the button reads **Accept**; after accepting, the section reads **Accepted by You**.
5. As **host/governor**: badge reads **Captain Accepted**, button reads **Approve**.
6. As **donor**, a pending request shows **Awaiting captain**.

## Checklist


adminmfl (approved):
Approved.

The overlap is the right thing to have surfaced before building rather than after. Documenting it at the function and pinning it with a test means the next person reads it as a decision with a named owner, not a hole someone missed — and the note about the correct fix (freeze the baseline once a final weight exists) means nobody re-narrows the window by reflex.

Rewriting the initial-window suite rather than bolting new cases on was also correct; leaving the old assertions would have left tests passing against a rule we no longer have.

Good call verifying awards day needed no change at all instead of building to the request as stated.

